PR #26, opened 2016-12-15
set docker host port to 80

hardfire:
This will ensure that the nginx app will listen on port 80, rather than a random host port.
I am not sure if this was intended but was confusing to me.


lalop:
thank you @hardfire for your pr.
I see this docker-compose as an helper to test the application in local, so the fact that the port is randomly choosen is a choice.
You can't assume that the end-user want to start the app on port 80

PR #27, opened 2016-12-17
Already in env

pierreozoux:
I tested and indeed, it is not needed.

### DIFF
--- a/app-entrypoint.sh
+++ b/app-entrypoint.sh
@@ -9,9 +9,6 @@ if [ "$DB_USERNAME" = 'root' ]; then
 	: ${DB_PASSWORD:=$MYSQL_ENV_MYSQL_ROOT_PASSWORD}
 fi
 
-echo "DB_USERNAME=$DB_USERNAME" >> .env
-echo "DB_PASSWORD=$DB_PASSWORD" >> .env
-
 if [ ! -d /var/www/app/storage ]; then
 	cp -Rp /var/www/app/docker-backup-storage /var/www/app/storage
 else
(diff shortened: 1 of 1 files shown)


lalop:
I think it's not needed until this variables changes, not ?

pierreozoux:
If $DB_USERNAME and $DB_PASSWORD are in env, then invoiceninja would recognize them well.

lalop:
Invoice ninja will take this one before the ones in .env ?

pierreozoux:
This I don't know, but what I know is that in the current setup, `.env` gets recycled with the container.

Should `.env` be persisted? (not in the README).

lalop:
`.env` as not to be persisted, all you need is to inject the good environment into your container.
If all this environment variable are set you shouldn't need to persist `.env` https://github.com/invoiceninja/dockerfiles/blob/master/app-entrypoint.sh#L12-L21

lalop:
@pierreozoux can we close this ?

pierreozoux:
The thing is that it is really strange that Invoice Ninja write things there, and expect to persist...
This is for me a bug, antipattern, and it already beat me:

For instance, in the emails sent, the link to the invoices where malformed.
Eventhough, the APP_URL was well filled in Settings.
But realised, that the app was writing this configuration to the `.env`, and guess what, after a restart, this conf was gone. So to persist it, I added: `APP_URL=https://invoice.indie.host/` to my env.

I think this is couterintuitive. But well, I don't have much time anymore, so I'll just close this. until somebody else will loose some time here again...
I think the good solution would be to read env, and give the precedence to env.
Then, put the rest in a conf file, and persist this conf for docker to not get recycled.

Issue #29, opened 2017-01-20
Cron does not send emails/reminders


WesleyCharlesBlake:
I've noticed that using the example docker compose file, the cron container starts and runs, however the emails and reminders are not sent.

I have configured my instances to send through Mailgun, and manual sends from the app itself are working correctly.

using v.2.8.0 of Invoice Ninja. debug doesnt show anything either.

Any ideas where I can start to look?

WesleyCharlesBlake:
Running the commands manually in the container work as well:

```
www-data@8d4e982a6ad5:~/app$ DB_USERNAME=root DB_PASSWORD=pwd /usr/local/bin/php /var/www/app/artisan ninja:send-reminders
2017-01-20 Running SendReminders...
1 accounts found
Strato Technology | DevOps as a Service: 0 invoices found
Done
www-data@8d4e982a6ad5:~/app$ DB_USERNAME=root DB_PASSWORD=pwd /usr/local/bin/php /var/www/app/artisan ninja:send-invoices
2017-01-20 Running SendRecurringInvoices...
2 recurring invoice(s) found
Processing Invoice 2 - Should send NO
Processing Invoice 3 - Should send NO
0 due recurring invoice instance(s) found
Done
```

WesleyCharlesBlake:
The logs indicate the following only
```
[2017-01-09 13:06:12] production.ERROR: Cache for currencies is not set  {"context":"PHP","user_id":0,"account_id":0,"user_name":"","method":"GET","url":"http://localhost","user_agent":"","ip":"127.0.0.1","count":1} []
[...]

PR #30, opened 2017-01-20
include app volume so cron can work

WesleyCharlesBlake:
This should fix issue #29

### DIFF
--- a/example/docker-compose.yml
+++ b/example/docker-compose.yml
@@ -29,6 +29,8 @@ services:
 
   cron:
     image: invoiceninja/invoiceninja
+    volumes_from:
+      - app
     links:
       - db:mysql
     user: www-data


lalop:
Hello @WesleyCharlesBlake 
Is this fixing the described issue ?
Mounting volume from app should only mount `/var/www/app/public` I think.

WesleyCharlesBlake:
hey @lalop ,

So in my dev env, I was able to get the jobs working, but in prod I see that my fix is not solving the situation.

I see the entry point on the cron container is:

```
    entrypoint: |
      bash -c 'bash -s <<EOF
      trap "break;exit" SIGHUP SIGINT SIGTERM
      sleep 300s
      while /bin/true; do
        DB_USERNAME=root DB_PASSWORD=pwd /usr/local/bin/php /var/www/app/artisan ninja:send-invoices
        DB_USERNAME=root DB_PASSWORD=pwd /usr/local/bin/php /var/www/app/artisan ninja:send-reminders
        sleep 5s
      done
      EOF'```

By tweaking the parameters here, I was able to get the jobs to run, but it does not seem to be working on the default config. Im still trying to figure out exactly where the problem is

PR #33, opened 2017-01-31
Issue #1311 "Upgrading issues 3.0.2"

freskimo:
When the running docker container is upgraded to a new version of invoice ninja, the `/app/public` directory is not updated with the new files because it is a volume. I also noticed that the `/app/storage` directory is not persistence (logs). This PR contains the fix by copying the new versions of the storage and public directory to a new location, then **rsync** is used to sync with the already persistent directories.

Furthermore, I prefer to have the `docker-compose.yml` in the top directory and changed some things there too (support for .env variable, named volumes & removed the "link" entries)

~~note: is there a specific reason why the db is tested during start (in `app-entrypoint.sh`)? This has no real use and adds on code. I would opt to leave it out but left it in this current PR.~~

edit 1: maybe the readme should reflect the changes better. e.g. mv example.env .env for docker-compose support.
~~edit 2: just noticed that I forgot the cron process in docker-compose.yml~~


Dockerfile
```diff
@@ -30,19 +30,20 @@ ENV INVOICENINJA_VERSION 3.0.3
...
+    && cp -r /var/www/invoiceninja-${INVOICENINJA_VERSION}/public /var/www/app/docker-new-public \
+    && rm -rf /var/www/invoiceninja-${INVOICENINJA_VERSION} \
     && chown -R www-data:www-data /var/www/app \
     && composer install --working-dir /var/www/app -o --no-dev --no-interaction --no-progress \
-    && chown -R www-data:www-data /var/www/app/bootstrap/cache \
-    # && echo ${INVOICENINJA_VERSION} > /var/www/app/storage/version.txt \
-    && mv /var/www/app/storage /var/www/app/docker-backup-storage \
-    && mv /var/www/app/public/logo /var/www/app/docker-backup-public-logo
+    && chown -R www-data:www-data /var/www/app/bootstrap/cache #\
+    # && echo ${INVOICENINJA_VERSION} > /var/www/app/storage/version.txt
```
lalop: Why did you change this lines to do the same things in fact ?

Dockerfile
```diff
@@ -52,7 +53,7 @@ ENV SELF_UPDATER_SOURCE ''
 
 
 #use to be mounted into nginx for exemple
-VOLUME /var/www/app/public
+VOLUME ["/var/www/app/public","/var/www/app/storage"]
```
lalop: Why do you add `storage` to volumes ?

lalop:
Your change are too different from the origine, can you make a simple pr only with what is needed please ?

freskimo:
Okay i will split up the different requests so it would be easier to discuss the changes.

lalop:
I start a new issue which is I think linked to this https://github.com/invoiceninja/dockerfiles/issues/34

lalop:
this should be fix https://github.com/invoiceninja/dockerfiles/commit/c07242eb1c7231b4c39adb98eef6406bddcf239f

PR #35, opened 2017-02-01
Issue#1311

freskimo:
Fixed issues with the declared volumes:

Why do we declare a volume in docker? Because we want to persist that specific data during restarts of containers. The only data of invoice ninja that needs to be persistent are the `public/logo` & the `storage`.
The logo contains the logo and the storage contains log files we want to persist over restarts of the container.

The old code did the following:
`&& mv /var/www/invoiceninja-${INVOICENINJA_VERSION} /var/www/app \`
But because the directory `/var/www/app/public` is declared as a volume, the mv will not be executed on the directory so all files get update except for this specific directory. This happens when a container is restarted . This is no issue when starting up a **new container** because the new volumes will be created and the files will be moved correctly.

The suggested fix will treat the volumes correctly as persistent and copies over the directories of the volumes to a new location. Only with an entrypoint script volume files can be copied over so in the script the new files will be rsynced to the existing volumes. I used rsync because of the benefits but a simple copy can also be used.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ MAINTAINER Samuel Laulhau <sam@lalop.co>
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         libmcrypt-dev zlib1g-dev git libgmp-dev \
-        libfreetype6-dev libjpeg62-turbo-dev libpng12-dev \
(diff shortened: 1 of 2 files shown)


lalop:
this should be fixed https://github.com/invoiceninja/dockerfiles/commit/c07242eb1c7231b4c39adb98eef6406bddcf239f

PR #36, opened 2017-02-01
docker composer improvement with use of the syntax v2

freskimo:
This improvement uses the new env_file syntax which eliminates the need to set the env variables per service. Also the app-entrypoint contains duplicated code, there is no need to create a `.env` file because Laravel will get the variables anyway. This makes it much easier to declare all config settings in the `.env` file


lalop:
I close this pr since it's not up-to-date with the repository and it didn't seem to use syntax v2

PR #37, opened 2017-02-04
Clear the .env file on every startup

cinemast:
Currently the variables get appended to the file each time there is a restart. This commit fixes this issue.

### DIFF
--- a/app-entrypoint.sh
+++ b/app-entrypoint.sh
@@ -9,6 +9,8 @@ if [ "$DB_USERNAME" = 'root' ]; then
 	: ${DB_PASSWORD:=$MYSQL_ENV_MYSQL_ROOT_PASSWORD}
 fi
 
+echo "" > .env
+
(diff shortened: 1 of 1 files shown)


lalop:
This may have some unwanted effect.
I have to make some test before merge it

freskimo:
Actually writing the .env file is duplicate. In Laravel when the env variables are set, they automatically are loaded. No need for this extra .env file...

This also eliminates your problem @cinemast

lalop:
@freskimo yes you are right, writing to `.env` from the entrypoint could be usefull only for mysql automatic detection variable

cinemast:
I overlooked that completly. @freskimo is right. Removing this file could also avoid misunderstandings such as this one.

freskimo:
I made a PR with the fix: https://github.com/invoiceninja/dockerfiles/pull/40

cinemast:
closing this in favor of #40

PR #38, opened 2017-02-04
Remove duplicate ENV variables

cinemast:
Since the variables are passed through an .env file they no longer need to passed directly, right?


lalop:
exact, thank you

PR #40, opened 2017-02-07
Removal .env file use docker env instead

freskimo:
In response of https://github.com/invoiceninja/dockerfiles/pull/37:

Instead of creating a .env file via app-entrypoint.sh, the env variables are set through docker.
Also added some doc.


lalop:
thank you

Issue #42, opened 2017-02-07
Testing db connection in app-entrypoint.sh


freskimo:
I want to open the discussion: 

**why is the db testing mechanism necessary?**

AFAIK It is not the responsibility of the container to check the db connection. Instead the application itself should verify if there is a working db connection. This simplifies the `app-entrypoint.sh` and loses the inaccurate try wait mechanism.

lalop:
The main reason is because I would like to intialize the database on first launch because the first migration can cause timeout via web request... but I didn't did it yet.

freskimo:
This should happen in the application and not on the level of container...
I understand the problem but this is not the right approach. The fix is needed in the `invoiceninja/invoiceninja` git repo and not here.

A solution to the problem would be to split up the queries in chunks and run it in a queue.

lalop:
What is wrong in doing this at the level container ?

freskimo:
You trying to fix a problem by solving it on a different location. This has a bad smell.
Other people hosting invoice ninja w/o docker have the same problem and this should be addressed in the application itself not on another level.

It is not wrong but docker files & entrypoints should stay as simple/clean as possible w/o too much extras. Offcourse you can add db imports, db backups, healthchecks... but this adds up to the startup time of the container e.g. when running multiple instances of the invoiceninja container with loadbalancing and you move all instances to a new location, then they first need to make a db call to see if there actually is a db. This process should happen as fast as possible.

**In my understanding the docker way is to spin up a container as fast as possible**

freskimo:
The docker way would be the following:
initially
`docker run --rm invoiceninja/invoiceninja php artisan db:seed --force`
this results in a db seed and when this is finished the container will be removed, then:
`docker run invoiceninja/invoiceninja`
to run the app.

I am working on a better readme and will post it soon.

freskimo:
Is running db:seed sufficient? Or should migrate also be run?

freskimo:
Are these the commands that need to be run?
```
php artisan optimize --force
php artisan migrate --force
php artisan db:seed --force
```

lalop:
I think this is all, and for the update 
```
php artisan optimize --force
php artisan migrate
php artisan db:seed --class=UpdateSeeder
```
according to http://docs.invoiceninja.com/en/latest/update.html

hillelcoren:
Yes, those are the commands needed to initially setup the app. 

You can combine the last two: `php artisan migrate --seed`

[...]

PR #44, opened 2017-02-07
DB init mechanism

freskimo:
Added doc and small changes to the app-entrypoint to allow initializing the db through commandline instead of using the web installer.

in response of https://github.com/invoiceninja/dockerfiles/issues/42

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,8 +66,6 @@ VOLUME /var/www/app/public
 
 WORKDIR /var/www/app
 
-EXPOSE 80
-
 COPY app-entrypoint.sh /entrypoint.sh
 RUN chmod +x /entrypoint.sh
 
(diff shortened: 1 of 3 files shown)


lalop:
Why don't just add some instructions in the Readme to explain what to do to init the database :

```
docker run -it --rm invoiceninja/invoiceninja php artisan optimize --force
docker run -it --rm invoiceninja/invoiceninja php artisan migrate --seed --force
```

lalop:
And the same for update

freskimo:
Well you are right this only a small improvement for the UX. But in this way the user is not bothered with the internal structure and framework of the application. 

We can add more cmds in the future like interactive installation via commandline, update, activate license, clear cache, full backup db + storage, change app key + db in case of a breach, ...

lalop:
Sorry but I prefer to refer to invoiceninja command then wrapping them

Issue #41, opened 2017-02-07
bad practice .env


freskimo:
It is considered bad practice placing the .env file in the repo.
It is better to make an example.env and tell the users to copy this file to .env. This reduces the risk of people pushing their passwords on some repo and to make sure no .env variables are pushed, a gitignore file should exclude any .env file.

https://github.com/invoiceninja/dockerfiles/blob/master/example/.env

miend:
I opened PR #48 to address this, pending approval from @lalop.

PR #48, opened 2017-02-16
Issue #41: Users shouldn't commit .env files to the repo...

miend:
... so a .gitignore is added to make sure any file by the name '.env' gets ignored. The existing .env is renamed to example.env with a comment clarifying its use.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+# Ignore environment variables.
+.env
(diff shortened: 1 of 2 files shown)


lalop:
Can you add a notice in the Readme too to explain how to use the example ?

miend:
If I do that, I might as well add instructions for using compose as the recommended run method in general. I can do that if you think it'll be helpful.

lalop:
Compose is the recommanded way of testing this image according to the readme.
I just want the readme to be up-to-date is the way to boot this image.

Issue #23, opened 2016-12-08
Docker-compose example failing because of mcrypt_get_iv_size()


Chewie:
Hello, upon cloning the repository and running the docker-compose, deployment runs fine but access returns an internal server error because an mcrypt function is deprecated:

~~~
ErrorException in McryptEncrypter.php line 46: Function mcrypt_get_iv_size() is deprecated

    in McryptEncrypter.php line 46
    at HandleExceptions->handleError('8192', 'Function mcrypt_get_iv_size() is deprecated', '/var/www/app/vendor/laravel/framework/src/Illuminate/Encryption/McryptEncrypter.php', '46', array('key' => 'SomeRandomString', 'cipher' => 'rijndael-128')) in McryptEncrypter.php line 46
    at McryptEncrypter->__construct('SomeRandomString', 'rijndael-128') in compiled.php line 7827
    at EncryptionServiceProvider->getEncrypterForKeyAndCipher('SomeRandomString', 'rijndael-128') in compiled.php line 7819
    at EncryptionServiceProvider->Illuminate\Encryption\{closure}(object(Application), array()) in compiled.php line 1353
    at Container->build(object(Closure), array()) in compiled.php line 1309
    at Container->make('encrypter', array()) in compiled.php line 1874
    at Application->make('encrypter') in compiled.php line 1410
    at Container->resolveClass(object(ReflectionParameter)) in compiled.php line 1387
    at Container->getDependencies(array(object(ReflectionParameter)), array()) in compiled.php line 1373
    at Container->build('Illuminate\Cookie\Middleware\EncryptCookies', array()) in compiled.php line 1309
    at Container->make('Illuminate\Cookie\Middleware\EncryptCookies', array()) in compiled.php line 1874
    at Application->make('Illuminate\Cookie\Middleware\EncryptCookies') in compiled.php line 10016
    at Pipeline->Illuminate\Pipeline\{closure}(object(Request))
    at call_user_func(object(Closure), object(Request)) in Pipeline.php line 32
    at Pipeline->Illuminate\Routing\{closure}(object(Request)) in compiled.php line 3286
    at CheckForMaintenanceMode->handle(object(Request), object(Closure))
    at call_user_func_array(array(object(CheckForMaintenanceMode), 'handle'), array(object(Request), object(Closure))) in compiled.php line 10021
    at Pipeline->Illuminate\Pipeline\{closure}(object(Request))
[...]

PR #52, opened 2017-05-12
Update php:7.0-fpm to php:7.1-fpm

arubacao:
According to the current [`.travis.yml`](https://github.com/invoiceninja/invoiceninja/blob/master/.travis.yml#L10) we only test against php 7.1


lalop:
The image is freezed to the 7.0 due to this issue https://github.com/invoiceninja/dockerfiles/issues/23

I didn't take the time to solve it but I will be happy to merge your pr if this issue is solved

arubacao:
I've fixed the problem and tested it locally :)
I basically removed the mcrypt related files & configs since nothing is depending on it anymore.

lalop:
thank you

PR #57, opened 2017-10-11
Mount MySQL volume to make data persistent

pottink:
When shutting down docker to for example open an other project docker environment, the mounted volume makes sure the database is being persisted.


lalop:
thank you

PR #61, opened 2017-12-15
Fixed MySQL volume mount to be from root.

welkie:
I noticed that the example docker-compose.yml file has the MySQL volume mounted from the current directory on the host side. I'm assuming this is a mistake, since this isn't normally where you'd mount your persisted data when using Docker to host applications. The convention is to use a directory in the root directory, usually /var or /srv.

I changed it to use /srv/invoiceninja/data instead, with the invoiceninja directory to prevent it from clashing with other Docker hosted applications. An example of this convention used in the industry is with [GitLab's Docker instructions](https://docs.gitlab.com/omnibus/docker/).

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@ The easiest way to test Invoice Ninja with docker is by copying the example dire
 
 To make your data persistent, you have to mount `/var/www/app/public/logo` and `/var/www/app/storage`.
 
+The MySQL volume is already mounted to `/srv/invoiceninja/data` so to persist data stored in MySQL (ex. invoices).
+
 All the supported environment variable can be found here https://github.com/invoiceninja/invoiceninja/blob/master/.env.example
 
 
--- a/example/docker-compose.yml
+++ b/example/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       MYSQL_DATABASE: ninja
       MYSQL_ROOT_PASSWORD: pwd
     volumes:
-      - /srv/invoiceninja/mysql:/var/lib/mysql
+      - /srv/invoiceninja/data:/var/lib/mysql
 
   app:
     image: invoiceninja/invoiceninja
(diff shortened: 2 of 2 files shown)


welkie:
@lalop Looks like you're the maintainer here. Any chance of this being merged?

lalop:
thank you @welkie, the current docker-compose.yml file is designed to test invoice ninja locally the fastest possible.
I'm not sure that having some file stored outside of the demo's folder is a good option in this case.
Anyway maybe this design isn't good and the docker-compose.yml file should be as production ready as possible in this case this folder is not the only point to change.
Discussion is open :)

hillelcoren:
My vote is to make 'production ready' the default. 

We've had similar issues with the app code in the past. We used to have APP_DEBUG set to true by default in the .env.example file (it helped debug problems when setting up the app) but found some users wouldn't notice and would run the app in production with it enabled. My takeaway lesson was it's better to make things a bit harder up front if it can prevent mistakes by beginners.

welkie:
@lalop I never thought of it from that perspective. Now it makes sense why it was set up this way. Myself, I only use Docker for the app itself in production, not for development. Perhaps ```example-dev``` and ```example-prod``` versions?

lalop:
I think I will update the docker-compose to be production as ready as possible and add information in the readme to people who want to test invoiceninja.

@welkie can you add some information in the readme to explain why the volume is mounted in that way ?

welkie:
@lalop I added this to the readme, but I noticed that the readme mentioned other volumes to be mounted from root to persist data:

> To make your data persistent, you have to mount `/var/www/app/public/logo` and `/var/www/app/storage`.

I'm not sure what data this is referring to, since I assume all invoice data is stored in MySQL. I want to get some clarification on this before this is merged. I'll also make them both mounted to `/var` or `/srv` for consistency before this is merged too.

lalop:
I rewrite the docker-compose file to be more production ready https://github.com/invoiceninja/dockerfiles/commit/13487557e6e63dc14f67e38656030ee5c9020c68
Thank you for your help

ps: any review of the readme is welcome

PR #64, opened 2018-01-28
[add] mysql performance improvement

AurelienLavorel:
Use docker volume to improve performance.

### DIFF
--- a/docker-compose/docker-compose.yml
+++ b/docker-compose/docker-compose.yml
@@ -6,7 +6,7 @@ services:
     env_file: .env
     restart: always
     volumes:
-      - ./srv/invoiceninja/mysql:/var/lib/mysql
+      - data-volume:/var/lib/mysql
 
   app:
     image: invoiceninja/invoiceninja
@@ -46,4 +46,7 @@ services:
         sleep 1h
       done
       EOF'
+
+volumes:
+  data-volume:
       


lalop:
can't we use docker's volume for invoiceninja's container too ?

AurelienLavorel:
Yes we can but it'll be harder to see logs from docker host.

lalop:
thank you

pierreozoux:
@AurelienLavorel what do you mean? Ah you men logs are in data folder?
grr :) stdout/stderr FTW \o/

But yes, agree with @lalop if we do it, then, let's do it for both.

AurelienLavorel:
Yeah I know, I'm not talking about access log.
`ls srv/invoiceninja/storage/logs/`
- failed-logins.log
- stacktrace.log

lalop:
humm, that 's embarassing this to log have their on logs wirter, they are ignoring the log setting https://github.com/invoiceninja/invoiceninja/blob/master/.env.example#L32
@hillelcoren can we make something for that ? Maybe log all in the error log with kind of label `[failed-login] ...` , `[stacktrace] ... `  ...

hillelcoren:
This is by design. Most people aren't using tools to analyze their logs, separating out the failed logins and stacktrace details makes it easier to see the login attempts and easier to read the errors in the logs.

That said we can support both behaviors, I'll look into it...

hillelcoren:
I understand the desire to keep the logs organized but I'm reluctant to change the format for existing users. Having the errors separated out makes debugging a lot easier for us. 

With that in mind... I've checked in a change to ensure the app respects the log setting for all but single. The other options are: daily, syslog, and errorlog.

Issue #71, opened 2018-05-03
Some small documentation issues with this repository


jbzdak:
I have recently used **modified** version of docker compose from this script to succesfully deploy invoice ninja for a client.

However, as a person without PHP experience, i have run into some issues: 

1. It appears that ``APP_KEY`` needs to conform to specified format (proper length, etc). It would be usefull to have this information in documentation here; 
2. Somehow storage folders had botched owner (root:root) while (at least partially) they need to be write-able by www-data (e.g. session storage); 

Anyways thanks for this repository it is great :)

lalop:
@jbzdak thank you for this comment.

@hillelcoren I'm right If I add in documentation that we can generate a key via the `key:generate` command ?
The complete command for us will be `docker run -e APP_CIPHER='AES-256-CBC' -it --rm invoiceninja/invoiceninja php artisan key:generate --show`

hillelcoren:
Do users see the /setup screen when they first launch the app?

When the setup form is submitted the app should automatically call key:generate

lalop:
In docker the environment variable is set via docker so the .env file shouldn't be updated by the app.

lalop:
Is that necessary to generate de APP_KEY ?

lalop:
Do you have any complete documentation on this kind of variable to add a link in the Readme ?

hillelcoren:
Sorry, I'm not sure I understand. 

Here's where PHANTOMJS_SECRET is mentioned in the docs: http://docs.invoiceninja.com/en/latest/configure.html#phantomjs

jbzdak:
I have added a PR that shows what updates would be helpful for me when I was deploying invoice ninja.

JamesK852:
This really helped me as I never used the php key:generate function before.

grahamPegNetwork:
This issue is over a year old, from what I can see it is impossible to get the docker file running without these instructions, please update the README with full instructions on how to get them running.

PR #72, opened 2018-05-14
Documentation clarification 

jbzdak:
Documentation that would have been helpfull for me when I was deploying invoice ninja in docker. 

See #71 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,11 +22,26 @@ docker run -d
   -e DB_DATABASE='ninja'
   -e DB_USERNAME='ninja'
   -e DB_PASSWORD='ninja'
+  -e PHANTOMJS_SECRET=`gVBUWzLjRWQq0lLqiO5RnGHvugUn+OuszNvymIiGex4jtrqznQa278sxHNNZZuhls54FhXWs9wFqN2pYmy8zKHEsMJoi60CSMZ6hkMiOymCf1IbHr2SxmidXeYyW+ZO6QjW6T2SwiJQH2fUFN82/yxtKHaDMz+7ilNQjZ0RGCQI=`
   -p '80:80'
   invoiceninja/invoiceninja
 ```
 A list of environment variables can be found [here](https://github.com/invoiceninja/invoiceninja/blob/master/.env.example)
 
+
+### Generating random variables
+
+Following variables should be random for your invoice ninja instance to be safe:
+
+* `APP_KEY` --- to generate app key run following command::
+        
+      docker run -e APP_CIPHER='AES-256-CBC' -it --rm invoiceninja/invoiceninja php artisan key:generate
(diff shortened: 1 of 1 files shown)


lalop:
Thank you for your PR.
I'm not sure that the phantomjs_secret is usefull in the Readme docker command, this command is a kind of easy way to start on the other side it could be a great improvement to add it into the docker-compose exemple that is more prod ready.
For the key generation, I would like to redirect on a common documentation for all the invoiceninja installs, what do you think @hillelcoren  ?

hillelcoren:
I think the 'Generating random variables' section is very helpful.

I'm not a fan of having a hardcoded PHANTOMJS_SECRET in case people end up using it, with the secret you can access an invoice without entering the client password.

README.md
```diff
@@ -22,12 +22,27 @@ docker run -d
   -e DB_DATABASE='ninja'
   -e DB_USERNAME='ninja'
   -e DB_PASSWORD='ninja'
+  -e PHANTOMJS_SECRET=`gVBUWzLjRWQq0lLqiO5RnGHvugUn+OuszNvymIiGex4jtrqznQa278sxHNNZZuhls54FhXWs9wFqN2pYmy8zKHEsMJoi60CSMZ6hkMiOymCf1IbHr2SxmidXeYyW+ZO6QjW6T2SwiJQH2fUFN82/yxtKHaDMz+7ilNQjZ0RGCQI=`
```
lalop: @jbzdak sorry for the long silence, as @hillelcoren mentined, can you please remove hard coded key and replace this by an explicit string ?

README.md
```diff
@@ -22,12 +22,27 @@ docker run -d
...
 A list of environment variables can be found [here](https://github.com/invoiceninja/invoiceninja/blob/master/.env.example)
 
 
+### Generating random variables
+
+Following variables should be random for your invoice ninja instance to be safe:
+
+* `APP_KEY` --- to generate app key run following command::
+        
+      docker run -e APP_CIPHER='AES-256-CBC' -it --rm invoiceninja/invoiceninja php artisan key:generate
```
lalop: you should add ` --show` to the command to avoid interaction

README.md
```diff
@@ -22,12 +22,27 @@ docker run -d
...
+   
+   Variable you need will be in the last line of output, in my case it was: `base64:WEMBumE5uARlC2PfrQ0WhcGN6lDTDwHubOGd3AyM7Ho=`
```
lalop: can you remove this one too ?

README.md
micah: Also, shouldn't these be enclosed in single quotes, and not backticks?

lalop:
closed in favor of #139

PR #75, opened 2018-06-21
Fix storage directory permissions

juanluisbaptiste:
Hi,

After completing the setup the application throws the following error:

```
(1/1) ErrorException
fopen(/var/www/app/storage/version.txt): failed to open stream: Permission denied
 
```
This seems to be because after _/var/www/app/docker-backup-storage/_ is copied to _/var/www/app/docker-storage/_, some files are created afterwards with root ownership inside that directory.

This PR fixes ownership after the storage directory has been copied.


juanluisbaptiste:
Sorry, mixed commits, will send it again.

PR #76, opened 2018-06-21
Fix storage directory permissions

juanluisbaptiste:
Hi,

After completing the setup the application throws the following error:

```
(1/1) ErrorException
fopen(/var/www/app/storage/version.txt): failed to open stream: Permission denied
 
```
This seems to be because after _/var/www/app/docker-backup-storage/_ is copied to _/var/www/app/docker-storage/_, some files are created afterwards with root ownership inside that directory.

This PR fixes ownership after the storage directory has been copied.
It also adds a .gitignore file, I hope you don't mind.


lalop:
Hey,
I tried to manage that by chowning all the directories https://github.com/invoiceninja/dockerfiles/blob/master/Dockerfile#L53
It seems that I missed something, can you give my the files that was created please ?

juanluisbaptiste:
The issue is not because you are missing fixing ownership of some files in the Dockerfile, the issue is that at some point when the container starts the directory /var/www/app/storage (which you have on the chown command in the Dockerfile) directory ownership gets changed back to root:

```
/var/www/app# ls -lth|grep storage
drwxr-xr-x   8 root     root     4.0K Jun 22 19:54 storage
```

Contents are correclty owned by www-data:www-data, but the storage directory does not so it cannnot write the version.txt file.

lalop:
I don't get that just after creating the container 

```
$ docker run --rm -it invoiceninja/invoiceninja ls -lth | grep storage
drwxr-xr-x   8 www-data www-data 4.0K Jun 25 07:36 storage
drwxrwxr-x   8 www-data www-data 4.0K Jun 10 10:42 docker-backup-storage
```

When did you get this error ?

juanluisbaptiste:
I launch the container using the included docker-compose file and I get the error of the first post, and storage is owned by root.

lalop:
I just tried again to reproduce this issue by doing that 

```
$ cd docker-compose
$ docker-compose down -v # warning this will remove container and volume
$ docker-compose up
$ # then setup app
```

But I don't get any permission error, can you give me a process to reproduce it.
I really would like to fix this issue.

Thank you

juanluisbaptiste:
```
# pull image
docker pull invoiceninja/invoiceninja

# use docker-compose.yml of the repository
cd docker-compose

# Edit env file and set DB parameters
vim .env

# delete volumes to be sure state is clear
rm -fr srv/

docker-compose up
```
Installer starts, click on submit and you will get the error:

`fopen(/var/www/app/storage/version.txt): failed to open stream: Permission denied`

lalop:
I just push a litte fix, pretty much the same as you but whith the logo dir too.

juanluisbaptiste:
Ok, I will test now...

juanluisbaptiste:
It worked.

lalop:
good, thank you for your time and help

PR #77, opened 2018-06-21
Create the database and database on start

juanluisbaptiste:
Hi, 

To avoid having the user manually create the database and database user on the database container, this PR adds code to _entrypoint.sh_ to automatically create the database and user needed to setup the application.

It uses the environment variables DB_HOST, DB_DATABASE, DB_USERNAME, etc to do the database setup.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ RUN apt-get update \
         zlib1g-dev git libgmp-dev unzip \
         libfreetype6-dev libjpeg62-turbo-dev libpng-dev \
         build-essential chrpath libssl-dev libxft-dev \
-        libfreetype6 libfontconfig1 libfontconfig1-dev \
+        libfreetype6 libfontconfig1 libfontconfig1-dev mysql-client \
     && ln -s /usr/include/x86_64-linux-gnu/gmp.h /usr/local/include/ \
     && docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ \
     && docker-php-ext-configure gmp \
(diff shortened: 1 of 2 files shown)


lalop:
Thank you for your pr but I don't think that is really interesting since the mysql image creates the user and database when this information are given at the container's creation https://hub.docker.com/_/mysql/

juanluisbaptiste:
Ah, I did't know that. Then at least the README should be updated to mention that MYSQL_USER and MYSQL_PASSWORD should be set.

juanluisbaptiste:
Adding these env vars to .env.example would be of great help:

```
MYSQL_ROOT_PASSWORD=changeme
MYSQL_USER=${DB_USERNAME}
MYSQL_PASSWORD=${DB_PASSWORD}
MYSQL_DATABASE=${DB_DATABASE}
```

That way only MYSQL_ROOT_PASSWORD from the mysql image needs to be directly set, the others are set from the invoice ninja env variables.

lalop:
I didn't know this syntax `${DB_USERNAME}`, is that working ?

juanluisbaptiste:
Yes, test it out !

lalop:
Just tried this notation, it doesn't seems to work for me in docker-compose, with version are you using ?

juanluisbaptiste:
You are right, it isn't working. I saw it on another project, tested it and for some reason it seemed to work but it doesn't, so forget I mentioned it.

I think adding MYSQL_ROOT_PASSWORD, MYSQL_USER, MYSQL_PASSWORD and MYSQL_DATABASE to the example .env file would suffice to make it clearer about the db configuration.

codedge:
@lalop This seems like a stale PR and can probably be closed?

lalop:
I think too

PR #91, opened 2019-02-27
Change Dockerfiles

Smolevich:
- Remove section MAINTAINER as deperecated
- ADD ARG for official php image tag
- Move all dependencies for build in env variable
- Update docker-compose.yml

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,37 @@
-FROM php:7.2-fpm
+ARG PHP_IMAGE_TAG=7.2-fpm
+FROM php:${PHP_IMAGE_TAG}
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,22 +1,37 @@
-FROM php:7.2-fpm
+ARG PHP_IMAGE_TAG=7.2-fpm
+FROM php:${PHP_IMAGE_TAG}
```
lalop: I wasn't oware of this way of doing things, is that permit to be overridden by a child image ?

Smolevich: I didn't understand your question completely. This way can used for building images with different php base images, you can set `--build-arg PHP_IMAGE_TAG=7.3-fpm` and it works.

PR #103, opened 2019-07-05
Use the correct docker host for mysql

gempain:
`mysql` is not a valid host. The container name is `db` as specified by the `docker-compose.yml`.


lalop:
good point, thank you

PR #106, opened 2019-08-11
Unzip quietly

djmaze:
This improves build time about 1/3 in my setup.


djmaze:
Well, looking at my results again, that was not quite right. In my CI setup builds on particular nodes take hours because the masses of output seem to slow down the build. Adding the `-q` switch makes those builds finish in just a few minutes. 

The verbose output probably does not bring any meaningful insights, so IMHO this change makes sense either way.

lalop:
thank you

PR #118, opened 2019-12-31
Update README.md

Justintime50:
Clarified some language and swapped the reference of php to 7.2 to reflect the actual image.


lalop:
thank you

PR #139, opened 2020-05-09
Update documentation for secret keys

### LINKED ISSUE #71: Some small documentation issues with this repository

### LINKED ISSUE #123: What's the purpose of APP_KEY?
Document this stuff please.

### LINKED ISSUE #127: [Question] Backup guidance
Hi @lalop please tell me if you agree with this procedure for backup, restore and update?

backup:
- make a sql dump
- copy files from docker volumes (but which ones - just `/var/www/app/public/`?)

restore:
- create new database container, seed it with the sql dump
- copy/overwrite files to docker volumes, and set owner (`www-root:www-root``) and permissions

update:
- update image version number in `docker-compose.yml` file

Am I missing something? This is standard for any docker app, but I've experienced errors from time to time when apps need "something extra", like running extra commands for pruning, caching, etc.

If you agree maybe we could put this in the readme?

codedge:
Fix #71, fix #127, fix #123

* Update readme with details how to generate secret keys - application key and phantomjs
* Integrated mounting persistence folder into docker command

Check if #72 can be closed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,8 @@ openssl rand -base64 10 | md5 |head -c10;echo
 head /dev/urandom | tr -dc A-Za-z0-9 | head -c 10 ; echo ''
 ```
 
+For further configuration and toubleshotting regarding PhantomJS and Invoice Ninja [see documentation here](https://docs.invoiceninja.com/configure.html?#phantomjs).
+
 ## Usage
 
 To run it:
--- a/README.md
+++ b/README.md
@@ -1,21 +1,59 @@
-DockerFile for invoice ninja (https://www.invoiceninja.com/)
+# Docker for [Invoice Ninja](https://www.invoiceninja.com/) 
 
 This image is based on `php:7.2-fpm` official version.
 
-To make your data persistent, you have to mount `/var/www/app/public/logo` and `/var/www/app/storage`.
+## Prerequisites
 
+### Generate an application key
 
-### Usage
+Before starting Invoice Ninja via Docker make sure you generate a valid application key. If you are not sure what an application key is, please visit [this blog post](https://tighten.co/blog/app-key-and-you/).  
+
+To generate an application just run
+
+```shell
+docker run --rm -it invoiceninja/invoiceninja php artisan key:generate --show
+```
+
+This will generate an application key for you which you need later.
+
+### Create folders for data persistence
+
+To make your data persistent, you have to mount `public` and `storage` from your host to your containers.
+
+1. Create two folder on your host, e. g. `/var/invoiceninja/public` and `/var/invoiceninja/storage`
+2. Mount these folders into your container - see [usage](#usage)
+
+You can create these folders wherever you want on your host system.
+
+### Generate a PhantomJS key
+
+If you use PhantomJS Cloud make sure you generated a secret key before starting Invoice Ninja. The following snippets will generate 10 char random keys.
+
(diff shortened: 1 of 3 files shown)


hillelcoren:
Thanks for the changes!!

One suggestion.. we may want to set the default key for PhantomJS Cloud to 'a-demo-key-with-low-quota-per-ip-address', I believe it supports 100 requests per day which is enough for most users.

codedge:
You are completely right - I just saw it in the [PhantomJS docs](https://phantomjscloud.com/docs/). I updated the readme.

README.md
```diff
@@ -1,20 +1,51 @@
...
+To make your data persistent, you have to mount `public` and `storage` from your host to your containers.
+
+1. Create two folder on your host, e. g. `/var/invoiceninja/public` and `/var/invoiceninja/storage`
+2. Mount these folders into your container - see [usage](#usage)
+
+You can create these folders wherever you want on your host system.
+
+### PhantomJS key
+
+The PhantomJS key is set to `a-demo-key-with-low-quota-per-ip-address`. This demo key is limited to 100 requests per day.
```
lalop, line 30: This doesn't seem to be really right.
Maybe we should be more clear, there is no valid phamtonjs key unless you set one.
Without valid key you will not be able to use the phatomjs service, which is not required by the way

codedge: Right, in the IN [`.env.example`](https://github.com/invoiceninja/invoiceninja/blob/master/.env.example#L29) it is set to

```
PHANTOMJS_CLOUD_KEY='a-demo-key-with-low-quota-per-ip-address'
```

but here in the Docker files we just `touch` the `/var/www/app/.env` file. 

So either we copy `.env.example` to `.env` or we pass this value in via environment var into the Docker container.
I personally would rely on the first version and would only override vars in via `-e VARNAME=KEY` that really need to be changed. That way we would work with reasonable set defaults in the IN project.

lalop: On the current docker setup we can't rely on `.env`, unless it's mounted in the container this file can disappear on any update. The way to set an environment var is definitively via docker.

I didn't knew this phantomjs key sorry for my last comment.

Anyway this key is not defined as a default value https://github.com/invoiceninja/invoiceninja/blob/master/config/pdf.php#L40 it's defined as an exemple https://github.com/invoiceninja/invoiceninja/blob/master/.env.example#L29 and this is not used in the image, so finally this var isn't set at all, This sounds good since the debian version come with phantomjs.

An other point that can be problematic for mainteners, is to keep pure Invoice Ninja information like that on the docker image readme, if there is any change in the app you have to report them here.

Last of all, phantomjs is deprecated and will be removed in next version.
So you may delete this section for me

codedge: That's why I copy this `.env.example` to `.env` here: https://github.com/invoiceninja/dockerfiles/pull/139/commits/73b9df53deaba3d7beaf550924885105f264d013

lalop:
thank you @codedge

Issue #119, opened 2020-01-08
Dockerfile doesn't copy things over properly


smcpeck:
Trying to use the docker image, but can't get things to populate the `/var/www/app/` directory properly.

It looks, to me, that the following line in the Dockerfile: `&& mv /var/www/ninja /var/www/app  \` should actually be `&& mv /var/www/ninja/* /var/www/app  \` as the way it currently is just results in a folder being created as `/var/www/app/ninja`.

Happy to do a PR and fix this, but just wanted to start the discussion here in case I'm missing something.

lalop:
If you run `docker run --rm invoiceninja/invoiceninja ls -all /var/www/app` you only see a folder ninja ?

smcpeck:
Well, the Dockerfile cleans up after itself further down, so it's unclear what's going on exactly.  I've moved to a custom/generic php container to just run the code from a bind mount, so I've gotten passed this issue, but feel it is probably a stumbling block still with the official image out of the box.

codedge:
I think this is solved. If you @smcpeck still have issues with the current version please let us know in a separate issue.
When using IN please stick to the `alpine` version.

PR #140, opened 2020-05-10
Improvements for Alpine image

### LINKED ISSUE #99: Don't run InvoiceNinja as root user
It would be nice if the Dockerfiles would use non privileged users instead of root where possible.

For example [like described here](https://medium.com/@mccode/processes-in-containers-should-not-run-as-root-2feae3f0df3b).

### LINKED ISSUE #102: alpine image: `Class LaravelCrawlerDetect does not exist`
I'm trying to setup InvoiceNinja with Docker (`invoiceninja:alpine` to be more specific) but I'm running in to this exception instead:

Screenshot_20190611_110526

I found [this forum thread](https://www.invoiceninja.com/forums/topic/class-laravelcrawlerdetect-does-not-exist/) but the hints in it doesn't help in my case.

Calling `php artisan optimize` within the container showed
```
Generating optimized class loader
The compiled services file has been removed.
```

but the error still appears (calling `php artisan clear-compiled` before didn't make a difference).
I tried to call `composer install` but it doesn't seem to work within the container. Should it usually be available in container? Or do I need to call it differently here (a specific path or something like that)?

I didn't try to change the source code tough because I thought if must be something else if it is working everywhere else.

The directory `bootstrap/cache/` is empty (besides the `.gitignore`).

Command `$ docker logs invoiceninja-app` only shows `11/Jun/2019:09:24:02 +0000 "GET /index.php" 500` for any request.

Do you have any advice how to debug this or even an idea for a bugfix?

It might be related to [#2598](https://github.com/invoiceninja/invoiceninja/issues/2598) or [#1071](https://github.com/invoiceninja/invoiceninja/issues/1071) but I would guess that tooling and dependencies should be properly baked in the Docker image, right?

**UPDATE:** this error does only appear with the Alpine image!

### LINKED ISSUE #126: [Bug] Alpine image doesn't set ownership for /var/www/app/
I struggled to get the alpine image to start. I discovered it does not set ownership for the files.

This is in the [debian image](https://github.com/invoiceninja/dockerfiles/blob/4.5.17/Dockerfile#L67):
````Dockerfile
chown -R www-data:www-data /var/www/app/storage /var/www/app/bootstrap /var/www/app/public/logo /var/www/app/.env /var/www/app/docker-backup-storage /var/www/app/docker-backup-public
````

The [alpine image](https://github.com/invoiceninja/dockerfiles/blob/4.5.17/alpine/Dockerfile#L49) is missing that line.

Also, surely it would be better to do this: `chown -R www-data:www-data /var/www/app/` instead of the above, since it is the parent of all those subdirectories?

codedge:
This PR introduces some bigger changes for the docker `alpine` image.

**Build image via `Makefile`**

This is some first step towards automated build. The build command for the `alpine` based image is encapsulated into a _Makefile_ and can be called via `make build-alpine TAG="<IN_VERSION>"` from the project root dir.

**Move php config into separate file**

The configuration for php has been moved into a separate file, split into `cli` and `non-cli` version. This file can easily be copied into the docker images  and do not clutter the Dockerfile itself.

**Run as non-root**

Fix #99 

IN is run as a separate user called `invoiceninja` inside the docker image. All files are `chown`ed to this user, some have also `www-data` as group to be writeable by the web server.

As this has some implications for the mounted folders `public` and `storage`, I still need to update the readme. I am just waiting to get #118 merged in.
// ping @lalop 😄 

**Ownership for `/var/www/app`**

Fix #126 

The folder `/var/www/app` is properly owned by the new `invoiceninja` user

**Various other alpine-related issues**

- Resolve #102: I cannot find any error in the current `alpine` version running with 4.5.18.
- Issue #119: I cannot confirm this problem in the `alpine` image.

I built the image on Linux (Debian 10) and MacOSX Catalina 10.15.4 without problems and run it currently for own business.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,16 +38,8 @@ RUN curl -o ${PHANTOMJS}.tar.bz2 -SL https://bitbucket.org/ariya/phantomjs/downl
     && ln -sf /usr/local/share/${PHANTOMJS}/bin/phantomjs /usr/local/bin \
     && rm -rf /var/lib/apt/lists/*
 
-# set recommended PHP.ini settings
-# see https://secure.php.net/manual/en/opcache.installation.php
-RUN { \
-		echo 'opcache.memory_consumption=128'; \
-		echo 'opcache.interned_strings_buffer=8'; \
-		echo 'opcache.max_accelerated_files=4000'; \
-		echo 'opcache.revalidate_freq=60'; \
-		echo 'opcache.fast_shutdown=1'; \
-		echo 'opcache.enable_cli=1'; \
-} > /usr/local/etc/php/conf.d/opcache-recommended.ini
+COPY ./config/php.ini /usr/local/etc/php/php.ini
+COPY ./config/php-cli.ini /usr/local/etc/php/php-cli.ini
 
 #####
 # DOWNLOAD AND INSTALL INVOICE NINJA
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,19 @@
+ifndef TAG
+$(error The TAG variable is missing.)
+endif
+
+# Docker Hub namespace
+HUB_NAMESPACE="codedge"
+
+# Image name
+IMAGE="invoiceninja"
+
+
+# Building the IN based on alpine image, tag the image.
+# The 'latest' tag is also being assigned to this image.
+.PHONY: build-alpine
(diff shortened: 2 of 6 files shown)


lalop:
And again thank you @codedge :D

PR #142, opened 2020-05-11
Explanation how to view log files

### LINKED ISSUE #137: Empty logs
Has anyone here managed to get logs in `/var/www/app/storage/logs`? That directory is always empty for me.

- the container's php-fpm master processes run as root
- the container's pool ("www") processes run as www-data or 82
- the container's log directory is `www-data:www-data` and `775`

codedge:
Fix #137

* Better description how to find log files
  * Show the users how view logs w/ `docker-compose logs` or `docker logs`
* Change default log to file but can be overridden by env var

⚠️ ~This depends on a change in #139, as only if this gets merged, the values from `.env.example` are used when building the containers.~

**Background of this PR**

In the current situation the user needs to know how to view log files from a docker container, as the log driver is set to `errorlog`.
The user must be proficient with Docker to run either `docker logs` or `docker-compose logs`. 

The issue #137 showed, that neither all users are proficient, nor do they conclude that when in Docker context, there is probably no log file. No, they are still looking for a file.

[**Example 1**](https://github.com/invoiceninja/dockerfiles/issues/137#issuecomment-626303684)
> [...]and were told to "check their log files".
> So I tried to check my log files and discovered I don't have any.

[**Example 2**](https://github.com/invoiceninja/dockerfiles/issues/137#issuecomment-626305119)
> If it is redirected to docker's output / error, I assume output would be redirected to /dev/stdout and errors to /dev/stderr?
> Put another way... Do you get logs in your container? Is it just me?

So this PR should add more documentation, which it does, and it should set a default, that users probably embrace more - nonetheless it gives enough configuration for proficient users to configure the application as they like.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,6 @@ docker run -d \
   -e DB_DATABASE='ninja' \
   -e DB_USERNAME='ninja' \
   -e DB_PASSWORD='ninja' \
-  -e LOG=errorlog \
   -p '9000:9000' \
   invoiceninja/invoiceninja
 ```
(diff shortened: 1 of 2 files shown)


codedge:
@lalop I updated the PR to only describe how to check logs. Log driver is restored to `errorlog`.

lalop:
perfect, thank you

hillelcoren:
Great work, the docs are excellent!

PR #143, opened 2020-05-12
Add caddy webserver, remove separate cron container

### LINKED ISSUE #84: cannot connect to docker; not listening on port tcp/80 (only tcp6/80)
Hi,
Thanks for providing a docker file.

As explained in by programster in issue https://github.com/invoiceninja/invoiceninja/issues/1718 there is no way to connect to the nginx instance running inside docker.

Before the issue got closed the IO was asked (https://github.com/invoiceninja/invoiceninja/issues/1718#issuecomment-338502012) to post in this repo.

So far, this issue is still prevailing.

Any hints?

cheers

### LINKED ISSUE #141: Moving cronjobs into main docker image
I just came across that in the `docker-compose` setup there is a separate `cron` container to run the cronjob for sending _reminders_ and _invoices_.

@lalop  Is there any particular reason why the cronjob does only exist for `docker-compose` and not for php docker images? 
I would vote for moving the cron into the docker images itself. Cronjobs can easily be run by the `alpine` or `debian`-based images and not one need to take care about settuing them up.

This would bring two advantages:

* We are going to have proper cronjobs inside the php docker images, for non-`docker-compose`
* We would not need another container just for running cronjobs w/ `docker-compose`

codedge:
Fix #84, fix #141 

This PR introduces the change from _nginx_ to [_caddy_](https://caddyserver.com/) for the big advantage to provide HTTPS 🔒 for every IN installation (with docker-compose). Hoorray 🥳 

It simplifies configuration by relying on the the default values from `.env.example` and just overriding the necessary ones by passing them in via env vars.

Furthermore it removes cronjobs as a separate container for the docker-compose setup. Relying on starting the crons from the host system is documented.


lalop:
I manually merge this due to conflict.
Thank you

PR #144, opened 2020-05-17
Stackoverflow for getting help

codedge:
As most of the questions are not related to code changes and more or less need answer how to run IN in a docker environment of with the docker setup provided, it make sense people would post their questions on StackOverflow. 
I created a separate tag [`invoice-ninja`](https://stackoverflow.com/tags/invoice-ninja/) so we can easily spot the new questions.


lalop:
thank you

PR #145, opened 2020-05-20
Build docker images via Github actions

codedge:
This PR is going to introduce docker images build via Github actions. It adds a separate _Makefile_ which takes care of the different images, the building and tagging.

Whenever a new release is created the docker images are going to be built from `master` branch.

**Preparation**

Two secretes are needed to login to Docker Hub:
* `DOCKERHUB_USERNAME`
* `DOCKERHUB_PASSWORD`

These need to set up in the repository before.


hillelcoren:
Nice!!

README.md
```diff
@@ -161,8 +154,3 @@ environment:
 to your `docker-compose.yml`.
 
 This generated log file will only hold Invoice Ninja information.
-
-
-### Known issues
-
-Phantomjs doesn't work on linux alpine https://github.com/ariya/phantomjs/issues/14186.
```
lalop, line 168: why did you remove that ?

alpine/Dockerfile
```diff
@@ -1,69 +1,85 @@
...
-RUN docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ \
-    && docker-php-ext-configure gmp \
-    && docker-php-ext-install iconv mbstring pdo pdo_mysql zip gd gmp opcache \
-    && echo "php_admin_value[error_reporting] = E_ALL & ~E_NOTICE & ~E_WARNING & ~E_STRICT & ~E_DEPRECATED">>/usr/local/etc/php-fpm.d/www.conf
+WORKDIR /var/www/app
+
+COPY ./alpine/entrypoint.sh /usr/local/bin/docker-entrypoint
+RUN chmod +x /usr/local/bin/docker-entrypoint
+
+ENV PHANTOMJS="phantomjs-2.1.1-linux-x86_64"
```
lalop, line 16: It doesn't seem to be used actually, we may remove this line

alpine/entrypoint.sh
```diff
@@ -1,47 +1,44 @@
...
-		fi
-	done
-fi
 
-# compare public volume version with image version
-if [ ! -e /var/www/app/public/version ] || [ "$INVOICENINJA_VERSION" != "$(cat /var/www/app/public/version)" ]; then
-  echo 'clone public directory'
-  cp -Rp /var/www/app/docker-backup-public/* /var/www/app/public/
-  echo $INVOICENINJA_VERSION > /var/www/app/public/version
+if [ "$1" = 'php-fpm' ] || [ "$1" = 'bin/console' ]; then
```
lalop, line 10: why this condition ?

config/php/php-cli.ini
```diff
@@ -0,0 +1,18 @@
...
+
+opcache.enable_cli=1
+opcache.fast_shutdown=1
+opcache.memory_consumption=128
+opcache.interned_strings_buffer=8
+opcache.max_accelerated_files=4000
+opcache.revalidate_freq=60
+# http://symfony.com/doc/current/performance.html
+realpath_cache_size = 4096K
+realpath_cache_ttl = 600
```
lalop, line 14: php.ini can be load in cli context, that way you can avoid duplication.

debian/Dockerfile
```diff
@@ -1,11 +1,16 @@
-ARG PHP_IMAGE_TAG=7.2-fpm-stretch
-FROM php:${PHP_IMAGE_TAG}
+ARG PHP_VERSION=7.2
+ARG INVOICENINJA_VERSION
+
+FROM php:${PHP_VERSION}-fpm-stretch
 
 LABEL maintainer="Samuel Laulhau <sam@lalop.co>"
```
lalop, line 6: you can add your name here too ;)

debian/Dockerfile
```diff
@@ -1,11 +1,16 @@
...
 
 #####
 # SYSTEM REQUIREMENT
 #####
+COPY ./alpine/entrypoint.sh /usr/local/bin/docker-entrypoint
```
lalop, line 11: is that normal you copy the alpine entrypoint ?

codedge: Good catch. Fixed.

debian/Dockerfile
codedge: Fixed 😉

config/php/php-cli.ini
codedge: I saw this separation in other projects... in cli context the memory limit is set to `2G`. See line 16 in that file:
```php
memory_limit = 2G
```
So only on cli we have a different setting for `memory_limit`. For the normal usage we did not explicitly set something which then uses the default of `128M`.

alpine/Dockerfile
codedge: Fixed

README.md
codedge: I thought as we use phantomJS in alpine, see line 
```php
RUN cd /usr/share \
    && curl -s -L https://github.com/Overbryd/docker-phantomjs-alpine/releases/download/2.11/phantomjs-alpine-x86_64.tar.bz2 | tar xj \
    && ln -s /usr/share/phantomjs/phantomjs /usr/local/bin/phantomjs
```

it works - or doesn't it?

alpine/entrypoint.sh
codedge: Fixed. Some old check.

README.md
lalop: not last time I tried

codedge: As agreed, the local PhantomJS version is removed from `alpine` version. The cloud version is the preferred one.

PR #154, opened 2020-05-28
Docker secrets support

### LINKED ISSUE #125: [Feature Request] Improve security by using docker secrets
Please support "docker secrets" for passwords, API, keys, secrets, etc. Most major apps (and all databases) support this.

So instead of this:
````yml
version: "3.7"
services:

  mariadb:
    environment:
      MYSQL_PASSWORD: foo
    # ...etc.

  invoiceninja:
    environment:
      DB_PASSWORD: foo
      APP_KEY: bar
      MAIL_PASSWORD: baz
    # ...etc.
````

We can do this:
````yml
version: "3.7"
services:

  mariadb:
    secrets:
      - mariadb_password
    environment:
      MYSQL_PASSWORD_FILE: /run/secrets/invoiceninja_mariadb_password     # supported
    # ...etc.

  invoiceninja:
    secrets:
      - mariadb_password
      - app_key
      - mail_password
    environment:
      DB_PASSWORD_FILE: /run/secrets/invoiceninja_mariadb_password     # unsupported
      APP_KEY_FILE: /run/secrets/invoiceninja_app_key                  # unsupported
      MAIL_PASSWORD_FILE: /run/secrets/invoiceninja_mail_password      # unsupported
    # ...etc.

  secrets:
    mariadb_password:
      file: .invoiceninja_mariadb_password
    app_key:
      file: .invoiceninja_app_key
    mail_password:
      file: .invoiceninja_mail_password
````

This is more secure. Most secure database deployments do it like this, so invoiceninja is the weak link in the chain (as we need to put the password in plaintext).

Here are [the docs](https://docs.docker.com/compose/compose-file/#secrets-configuration-reference).

### LINKED ISSUE #149: Support `_FILE` env vars for secret configuration
Hi folks, thanks for the great work, I've got a favour to ask. 

I'm deploying `invoiceninja` on docker swarm and everything runs smoothly until I realise that I *must* provide a writable `/var/www/app/.env` file. Now, I use a [docker config](https://docs.docker.com/compose/compose-file/#configs) for this file which can never be writable (`tmpfs`, for the purposes of rollback). I can't specify secrets from the `environment` stanza because of [a Docker swarm design decision](https://docs.docker.com/engine/swarm/secrets/#how-docker-manages-secrets) soooo, I am wondering if it would be possible to follow the style of mariadb with their support of secrets:

> https://github.com/docker-library/docs/blob/master/mariadb/README.md#docker-secrets

They offer `_FILE` for the environment variables. So, concretely, you'd provide:

- `DB_PASSWORD` + `DB_PASSWORD_FILE`
- `APP_KEY` + `APP_KEY_FILE`
- `APP_SECRET` + `APP_SECRET_FILE`

This would allow me to specify a  `docker-compose.yml` which can be deployed on Docker swarm with:

```yaml
invoiceninja:
    environment:
        API_SECRET_FILE=/var/run/secrets/api-secret-file-v1
```

I guess you'd also need to change settings logic to not allow these to be changed if using the `_FILE` variables? Unsure. I don't know if you have many swarm deployments which might justify this work but this change would allow them. Otherwise, we have to store secrets in the `docker-compose.yml` configuration which is no bueno.

codedge:
Fix #149, fix #125 

This PR introduces the usage of docker secrets, better the usage for specifying a `_FILE` to read the secret from.

When using `docker-compose` you can define certain environment variable like this:
* `APP_KEY_FILE`
* `S3_KEY_FILE`
* etc.

You can find a full list of supported variables in the readme.

This change is backwards compatible and support the `alpine` version for 4.x and 5.x.


codedge:
Correct, the entrypoint script for `alpine` images, v4 or v5, is only a single script. It is fine for both version of IN.

decentral1se:
:partying_face: :partying_face: :partying_face:

lonix1:
+1 !

PR #156, opened 2020-05-29
Remove debian versions

codedge:
This PR removes the debian based docker containers. We're only focusing and supporting _Alpine_-based images.
Also the latest changes are only made for the _Alpine_-based versions.

The images are
* smaller
* less cluttered with stuff we don't need


lalop:
@codedge can you rebase the branch please ?

codedge:
I did a rebase. Does that help?

PR #161, opened 2020-06-05
Extracting version out of tag

codedge:
This PR makes it possible to update the Dockerfiles without increasing the version of IN.

**Use case**

Create a new release/tag `5.0.4` will built the image for IN 5.0.4.
Now, when something is wrong or needs to be fixed with the Dockerfiles, entrypoint, etc. we delete the release/tag and create it again to deploy changes to the docker setup.

But now you can just create a new release `5.0.4-p1` which will built the new images, with IN `5.0.4` but also with the changes made to the docker setup. 

So we suffix the IN version with `-pX`. And we do not need to delete and recreate releases here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,9 @@ IMAGE="invoiceninja"
 # Check if v5 tag is passed, so that a v5 version should be built
 IS_V5=$(shell echo ${TAG} | egrep ^5)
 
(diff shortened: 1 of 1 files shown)


lalop:
what id "p" for ?

codedge:
`p` is just for `patch`. It can be anything as long as there is a dash `-`. 
So we could have 

`5.0.4-p1`, `5.0.4-p2` ... patch 1, patch 2, it is just to give it some naming.

PR #191, opened 2020-08-19
Minor fix for docker composer

turbo124:
One small thing noticed is that the default db_name of db-ninja-01 doesn't seem to work, however replacing with ninja sees a smooth install.

Unsure of the root cause, but feel using ninja may be a better experience for users?


codedge (approved):

codedge:
Totally right. 👍

PR #258, opened 2021-01-10
Enhancement and Fixes for Dockerfile

lwj5:
Hi,

I've made multiple updates, and have split them up into multiple commits for easier review.

1. Update cron to depend on app 
   - This prevents both the cron and app to concurrently copy files from BAK + migration, *IF* heath check is set up

2. Move BAK after npm run prod
   - A mounted volume will not be updated if files are copied from `public` to `BAK_PUBLIC_PATH` before npm run prod is ran
   - remove PDO extension as it comes with fpm
   - Please help to check if the following three lines are legacy, IN works well without them
```
      # oniguruma-dev \
      # git \
      # busybox-suid \
```
   
3. Update Dockerfiles to use ENV for BAK paths
   - Allow updating of BAK paths using ARG
   - Export ARG to ENV

4. Update entrypoint.sh to use ENV paths + cleanup
   - Use BAK paths from ENV
   - streamline if else statements
   - replace some `cp -Rp` with `cp -a`, to reduce copying during `cp -au` when updated

5. Use uid for `USER` to allow running in k8s with non-root PSP
   - using `USER <name>` prevents running on a harden k8s cluster

The last 2 commits should be straightforward.
Please let me know if you have any concerns. Hope it helps!

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -1,4 +1,6 @@
 ARG PHP_VERSION=7.2
+ARG BAK_STORAGE_PATH=/var/www/app/docker-backup-storage/
+ARG BAK_PUBLIC_PATH=/var/www/app/docker-backup-public/
 
 FROM php:${PHP_VERSION}-fpm-alpine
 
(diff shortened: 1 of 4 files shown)


turbo124:
@lwj5 

Thank you for this! We really appreciate the help with our Dockerfile. 

If you can reverse the change that removed ```git``` as we use it to self update the application.

Thanks!

lwj5:
Reverted the change.

No problem, thanks for developing IN. Great stuff!

turbo124:
@lwj5 thanks again! if you have any suggestions for k8s implementation of InvoiceNinja we'd love to hear them!

PR #262, opened 2021-01-18
Use the schedule:work command directly instead of a bash script

tms0:
In Laravel 8.x, there is a new schedule command to run only one process in foreground instead of a cron job : https://laravel.com/docs/8.x/scheduling#running-the-scheduler-locally

They say it's for development purpose, but it seems perfect for Docker (one container = one process) and it can replace gracefully the cron.sh script.

What do you think about it ?


turbo124:
@tms0 thanks for this. If they tag this as production ready then we'll certainly consider it.

It does make me a little nervous as if the process dies then it has big implications.

tms0:
I understand, I added a "unless-stopped" line in my docker-compose to handle this case.

turbo124:
> I understand, I added a "unless-stopped" line in my docker-compose to handle this case.

Fair point. I'll watch how this progresses on Laravel to see if they consider this production ready.

turbo124:
@tms0 

I think we can merge this now. Are you able to add to to this PR the unless-stopped line in the docker-compose  please?

tms0:
Hi @turbo124,

I think I changed my mind ! I read a lot of things about how to manage cron things in Docker (one container vs multiple containers) and I am now convinced that one container with supervisord is not a bad solution (at least in my infrastructure).

There is a page in Docker docs about it : https://docs.docker.com/config/containers/multi-service_container/
And you can find an example on the nextcloud project : https://github.com/nextcloud/docker/tree/master/.examples/dockerfiles/cron

Also, this PR did not fix the queue issue for me, I still have to set the QUEUE_CONNECTION to "sync". I think a third container is required to run something like "php artisan queue:listen".

If you don't mind, I will open another PR with supervisord. Maybe in an "example" directory like in the nextcloud project ?

turbo124:
So you would use supervisord to run and monitor schedule:work?

tms0:
Yep, and also queue:listen

tms0:
Hi @turbo124,

Here's the PR with supervisor : https://github.com/invoiceninja/dockerfiles/pull/270

PR #270, opened 2021-01-27
Add example to run php-fpm, scheduler and queue worker in the same container with supervisor

tms0:
As I said before, I think it's not a bad thing to run related processes in the same container. It reduces the complexity and it uses significantly less RAM than 3 containers !

You can see the same thing in the Nextcloud projet : https://github.com/nextcloud/docker/tree/master/.examples/dockerfiles/cron
And the docker documentation about this subject : https://docs.docker.com/config/containers/multi-service_container/

How do you feel about that ?


turbo124:
@tms0 this looks good. I think we just need to expand the supervisor block with a bit more information. This is the laravel suggested block

```
process_name=%(program_name)s_%(process_num)02d
command=php queue:work database --sleep=3 --tries=3
autostart=true
autorestart=true
user=www-data
numprocs=2
redirect_stderr=true
stopwaitsecs=3600
```

tms0:
Hi @turbo124,

After reading the [queues documentation](https://lumen.laravel.com/docs/5.1/queues), I still don't know if we should use `queue:worker` or `queue:listen`, what do you advice ?

Regarding supervisord, I'm not sure about some options : 
  - autostart=true => it's the default value
  - autorestart=true => the default value "unexpected" seems better to me
  - user=www-data =>  I put "invoiceninja" in the supervisord section directly, so everything is running under this user

And for the "numprocs" option, maybe we can put this in an ENV var to allow users to adapt it depending on their needs ?

turbo124:
@tms0 thanks for the thought you have put into this PR! 

I'll merge and test. I think we'll also need to PR the changes to the docker-compose file to remove the additional container.

turbo124:
I changed queue:listen to queue:work as this is the recommended way in the laravel docs.

PR #279, opened 2021-01-28
Cleanup supervisor conf

tms0:
Hi @turbo124,

I made some minor cleanup :
- Removed user in the supervisord.conf since supervisor is already running with the low-privileged invoiceninja user
- Put pid in /tmp to save some "mkdir / chown" steps


turbo124:
@tms0 thanks for this, i'm going to wait a bit for our previous changes to bake for a while before adding more PRs. I'll loop back to this in a few releases time.

lwj5:
Hey there,

Just my 2 cents. I haven't tried this, but I don't think the queue worker should be running as `--daemon` as supervisor is supposed to manage the number of workers. Generally, running daemon would cause supervisor to keep spawning new processes as a daemon returns `exit 0` after creating a background process.

You may also want to clean up some redundant config:

- process_name can be removed for all except `program:queue-worker` as they are not needed
- `autostart=true` is by default and can be removed

Lastly, it's good practice to stop containers on a process failure, and allow restarting to be managed on a higher level, such as on docker-compose or k8s. Restarting on errors in docker containers could obscure underlying problems to these orchestration platforms.

To kill supervisor upon exit of other processes, in my work, we use a `shutdown.sh` file containing these lines.

```bash
#!/bin/bash

printf "READY\n";

while read line; do
  echo "Processing Event: $line" >&2;
  kill -SIGQUIT $PPID
done < /dev/stdin

```

and supervisor config with the following lines
```
[eventlistener:shutdown]
command=/shutdown.sh
events=PROCESS_STATE_STOPPED, PROCESS_STATE_EXITED, PROCESS_STATE_FATAL
stdout_logfile=/dev/stdout
stdout_logfile_maxbytes=0
stderr_logfile=/dev/stderr
stderr_logfile_maxbytes=0
```

tms0:
Another thing, you added directly "redis" as the queue driver in the supervisord configuration :
https://github.com/invoiceninja/dockerfiles/blob/84db77cde85814e02a3fa5344d786350726a66cd/config/supervisor/supervisord.conf#L39

I think we should let users to chose the driver (database by default)

turbo124:
oh! default should be database....

lwj5:
Hey there, would you be able to merge this?

I would like to make some changes on top of this

turbo124:
@lwj5 :)

PR #284, opened 2021-02-09
got all the vars onto the dotenv file

danielgroen:
also added a gitignore

the env file is now moved to a .env file since docker-compose only can read out a .env and not a env file.


turbo124:
Thanks again!

PR #288, opened 2021-02-13
now a working sample of .env and docker-compose

danielgroen:
Because my last pull request broke the Document downloads, i've created a new working .env file.

Although, it would be nice if the `CNAME` and `PORT` would be loaded in invoiceninja itself so the .env could be shorter.
u can combine the vars then like this:
`APP_URL=${PROTOCOL}${CNAME}:${PORT}`

i've also added container names. This is essential if you are running another nginx container, this will crash docker, so now they have their unique container names. check by: `docker ps -a`


turbo124:
@lwj5 are you able to review this for me please?

danielgroen:
so what i've changed are:

- env to .env since only then you can use .env vars in your docker-compose file
- removed the exposure of the ports of db since these are not needed to be exposed
- added the db .env vars into the db container so now you only have to declare them once instead of twice
- added container names, so you have less change to have duplicated container names if you have other docker container on your server
- set exposure port into the .env var since you don't want to block your port 80 always
- changed the network to 'default' so you don't have to declare the network
- put the default extra_host name into a CNAME var so nothing needs to be changed in the docker-compose file

danielgroen:
we can also change the invoiceninja version by .env file:
image
this will easily update you to the newest version of the newest image,

what about it @turbo124 & @lwj5

.env.example
```diff
@@ -1,7 +1,14 @@
-APP_URL=http://in.localhost:8003/
-APP_KEY=<insert your generated key in here>
+APP_URL=http://in.localhost:8004
```
lwj5, line 1: Any reason for the change in port?

docker-compose.yml
```diff
@@ -34,35 +36,40 @@ services:
...
   # THIS IS ONLY A VALID CONFIGURATION FOR IN 4. DO NOT USE FOR IN 5.
   # cron:
   #   image: invoiceninja/invoiceninja:alpine-4
   #   volumes:
-      # - ./docker/app/public:/var/www/app/public:rw,delegated
-      # - ./docker/app/storage:/var/www/app/storage:rw,delegated
-      # - ./docker/app/public/logo:/var/www/app/public/logo:rw,delegated
+  # - ./docker/app/public:/var/www/app/public:rw,delegated
+  # - ./docker/app/storage:/var/www/app/storage:rw,delegated
+  # - ./docker/app/public/logo:/var/www/app/public/logo:rw,delegated
```
lwj5, line 72: Could you add the correct spaces between the "#" and "-", so when you uncomment the whole block the YAML is correctly indented.

.env.example
```diff
@@ -1,7 +1,14 @@
...
+CNAME=in.localhost
+PORT=8004
+
+APP_KEY=docker run --rm -it invoiceninja/invoiceninja php artisan key:generate --show
```
lwj5, line 5: Let's not change this, I would imagine someone would use this as the actual app key. Rather and add a comment the line before.
```
# Run `docker run --rm -it invoiceninja/invoiceninja php artisan key:generate --show` and copy the output here
```

lwj5 (commented):
My apologies, I had been tied up for the past few weeks.

Moving env to .env.example is a really good idea. I've added some comments, the rest looks good!

Let's leave the version as it is for now. I can see it two ways.

1. Pro: you have more control over the version
1. Con: you wouldn't get the updated version if someone changes the .env.example on the git

It's a tough choice

.gitignore
```diff
@@ -0,0 +1,8 @@
+*.sql
+
+docker/mysql
+docker/app/*/
+!.keep
+
+.env
+!.env.example
```
byteSamurai, line 8: Maybe you could also add `docker-compose.override.yml` here please?

Imho it would be also a good idea to include all files starting with `.env.*` so it would be possible to create a `.env.production` and stay with the upstream

PR #291, opened 2021-03-02
auto db backup script

danielgroen:
Hi,

As promised, i'd do a pr for automated db backups.
It stores it in the `docker/mysql/bak` folder


turbo124:
Thanks for this! I'll merge but I'll modify this slightly so that the backups are an optin feature rather than default.

PR #300, opened 2021-03-24
Updated Dockerfile and supervisord.conf

lwj5:
Dockerfile_v5
- Moved things around in Dockerfile so that slower items are applied first. This allow us to use the build cache more effectively if only small files are changed (like supervisord.conf)
- Remove useless addgroup (only used in v4)

supervisord.conf
- Fixed shutdown.sh
- removed redundant autostart (default is already true)
- remove autorestart for php-fpm, it should be taken care of on a higher level (e.g. k8s, docker-compose)

entrypoint.sh
- wait for DB to start, if not migration will throw exception


turbo124:
Thank you for this,

one question, will removing the addgroup functionality break the docker-compose for V4 users?

PR #306, opened 2021-03-25
Reshuffle Dockerfile to optimise caching

### LINKED ISSUE #293: Improve caching of layers in Dockerfile to reduce container size
The Dockerfile uses a multi-stage build, which is good. But it pulls two 170MiB+ file layers (zip files I think), *every time*, which is not good.

Since there are multiple releases every day (which is cool, it means this project is moving fast!), every redeploy is **very** heavy. It will make people reluctant to try new versions, and complain about old bugs that are already fixed.

I want to help but I'm not a PHP guy, so I'm unsure if there's any way to minimize those zip files. Are there any PHP users here who can advise? The main thing is to place those stages toward the end, to benefit from docker's caching of layers. But the files themselves are huge, how can we split them into smaller bits?

lwj5:
- fixes #293


hillelcoren:
You are awesome, thank you!!

Issue #183, opened 2020-08-11
InvoiceNinja v5 on Raspberry Pi 4 (ARMv7)


kewngen:
There have been other mentions about IN on RPi, for example: #108 
I am still unable to get it to work. When I change the mysql image as mentioned in #108 I get the following error:
```
standard_init_linux.go:211: exec user process caused "exec format error"
```

And when I try to build the Dockerfile_5 I get:
```
URL: https://download.cypress.io/desktop/4.6.0?platform=linux&arch=ia32
Error: Failed downloading the Cypress binary.
Response code: 404
Response message: Not Found

----------

Platform: linux (Alpine Linux - 3.11.6)
Cypress Version: 4.6.0
[...]
The command '/bin/sh -c npm install' returned a non-zero code: 1
```

Anyone able to run IN v5 on a RPi 4? If so, please provide some instructions on how to make it work.

dimkasta:
Any progress with this? Currently I am getting 

```
standard_init_linux.go:219: exec user process caused: exec format error
```

pixlguru:
"Error: Failed downloading the Cypress binary."
cypress doesn't run on a raspberry pi because it is not available for _arm_

horschdel:
Any progress on getting Invoice Ninja to run in docker on Raspberry Pi 4?

vsisl:
I would also be interested in running Invoice Ninja in docker on RPi4.

lwj5:
Anyone interested in testing an ARM64 image?  Build passed but I’m not even sure whether it will work.

tacticz:
Hello,

It is very possible to have invoiceninja running on a Raspberry Pi 4, as I currently have it 😃 

### ARM64, ARMv8, aarch64
First it is perhaps important to specify that the Raspberry Pi 4 has an **ARM v8** based Cortex-A72 processor as is stated on the [official raspberrypi.org page](https://www.raspberrypi.org/products/raspberry-pi-4-model-b/specifications/), this is also the case for the Raspberry Pi 3 Model B, B+ and in later models of the Raspberry Pi 2 which are Cortex-A53 based ([see here](https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2837/README.md)).
Consequently, all of those models **_should_** be usable for what is described hereunder, although I personally only tested this on a Raspberry Pi 4 B.
The terms **ARM 64**, **ARM v8** and **aarch64** are used interchangeably to describe this processor architecture.

[...]

PR #307, opened 2021-03-26
Build on ARM64

lwj5:


tacticz:
Hello @lwj5,

First of all thank you for integrating the ARM64 build in the official docker image. I can confirm it is working on my RasPi 4!
 
In issue #183 you asked me whether I could help with the ARM64 version, and I told you I'd be glad to help, so here is my feedback:

Although **the dockerhub's invoiceninja/invoiceninja:5 image is now running on ARM64**, the current `docker-compose.yml` file is still using:
```
  db:
    image: mysql:5
```
Which, as explained under the **Modifying docker-compose.yml** section of [my post](https://github.com/invoiceninja/dockerfiles/issues/183#issuecomment-807280903) , **is NOT ARM64 compatible**, and outputs the following message when issuing `docker-compose up -d`:
```
5: Pulling from library/mysql
docker: no matching manifest for linux/arm64/v8 in the manifest list entries.
```

I had to change the image reference to:
```
  db:
    image: mariadb:10.4
```
Which **is ARM64 compatible** in order to be able to launch all 3 containers on the RasPi.

Maybe you could update the `docker-compose.yml` as follow to make things easy for people willing to deploy on ARM64:
```
  db:
    image: mysql:5
    # Use MariaDB when running on ARM64
    #image: mariadb:10.4
```
Hope this can help.

lwj5:
Thanks @tacticz! Good to hear. Can I check if the env for mariadb remains the same as mysql?

tacticz:
Absolutely!
I've checked both [MySQL](https://hub.docker.com/_/mysql)'s and [MariaDB](https://hub.docker.com/_/mariadb)'s dockerhub pages and I can confirm that both **Environment Variables** and **Docker Secrets** sections are a copy/paste of each others with only the term `mysql` replaced by `mariadb`.

The **only exception** being the **`MYSQL_ONETIME_PASSWORD`** which is **not available in MariaDB**.
> Sets root (not the user specified in MYSQL_USER!) user as expired once init is complete, forcing a password change on first login. Any non-empty value will activate this setting. NOTE: This feature is supported on MySQL 5.6+ only. Using this option on MySQL 5.5 will throw an appropriate error during initialization.

Which means that **environment variables set for MySQL can be reused _as is_ for MariaDB configuration**.

lwj5:
Thanks @tacticz for the info. 

I won’t be able to work on it this week. I’ll try to get on it the following week.

turbo124:
@tacticz if you PR your changes we can merge that in :)

PR #315, opened 2021-04-01
Added MariaDB to Docker compose for ARM64

tacticz:
Hello, following the suggestion made by @turbo124 in the [discussion](https://github.com/invoiceninja/dockerfiles/pull/307#issuecomment-810126351) regarding commit #307, I've created this PR which brings the following modifications:

#### General remark

My IDE (Atom) is configured to use the included `.editorconfig` file. Nevertheless it seems that the files I edited got modified lines which I didn't touch because:
 - trailing spaces were suppressed at the end of some lines
 - a newline was added at the end of the edited files

Consequently `diff` displays those modifications, making it less clear what lines were intentionally changed. 

#### MariaDB instead of MySQL on ARM64 (adae9a0)

Commented out lines in the `docker-compose.yml` file were added to suggests using `image: mariadb:10.4` instead of `image: mysql:5` when running on ARM64. This was tested and is working on a Raspberry Pi 4.

The same was added to the `config/mysql/Dockerfile` but wasn't tested yet, although I can't see a reason why it wouldn't behave as correctly as it does in the standard deployment.

The `README.md` file was also updated to include a hint on the change required for ARM64 architecture compatibility.\
(Also really minor corrections were made in the existing text 😉)

#### MYSQL environment variables moved to the `env` file (acdb708)

Since it is stated in the README that:
> Instead of defining our environment variables inside our docker-compose.yml file we now define this in the env file

It seemed logical to also move the `MYSQL_` related environment values to this `env` file, hope you will agree with this move.

#### Final remark: first launch database failure

What I could notice is that when **deploying for the very first time using `docker-compose up -d` a database failure is reported**:
> SQLSTATE[42S02]: Base table or view not found: 1146 Table 'ninja.accounts' doesn't exist (SQL: select * from `accounts` limit 1)

I could notice that, although the `docker/mysql/data/ninja/` and all other mysql related files get created, **the invoiceninja related database doesn't appear**.

Stopping the containers (`docker-compose down`) and restarting them (`docker-compose up -d`) fixes the problem as, this time, all invoiceninja related tables are correctly created. I haven't had the time to dig into this, and to be honest I don't even know where to start with that problem. I have the feeling this might be a timing problem, since during the very first run mysql has to initiate it's own file structures and databases, so it might be that the initial migration command from invoiceninja cannot execute because mysql itself isn't ready yet?

I simply wanted to let you know about this situation and maybe would it be adequate to create an issue about this in the repo?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,7 @@ services:
       - app
     # Run webserver nginx on port 80
     # Feel free to modify depending what port is already occupied
-    ports: 
+    ports:
       - "80:80"
       #- "443:443"
     networks:
@@ -32,20 +32,22 @@ services:
       - ./docker/app/storage:/var/www/app/storage:rw,delegated
     depends_on:
       - db
-    networks: 
-      - invoiceninja  
+    networks:
+      - invoiceninja
     extra_hosts:
       - "in5.localhost:192.168.0.124 " #host and ip
 
   db:
     image: mysql:5
+#    When running on ARM64 use MariaDB instead of MySQL
+#    image: mariadb:10.4
 #    For auto DB backups comment out image and use the build block below
 #    build:
 #      context: ./config/mysql
     ports:
       - "3305:3306"
     restart: always
-    environment: 
+    environment:
       - MYSQL_ROOT_PASSWORD=ninjaAdm1nPassword
       - MYSQL_USER=ninja
       - MYSQL_PASSWORD=ninja
@@ -82,7 +84,7 @@ services:
   #     EOF'
   #   networks:
   #     - invoiceninja
-  #     
+  #
 
 networks:
   invoiceninja:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -47,11 +47,7 @@ services:
     ports:
       - "3305:3306"
     restart: always
-    environment:
-      - MYSQL_ROOT_PASSWORD=ninjaAdm1nPassword
-      - MYSQL_USER=ninja
-      - MYSQL_PASSWORD=ninja
-      - MYSQL_DATABASE=ninja
+    env_file: env
     volumes:
       - ./docker/mysql/data:/var/lib/mysql:rw,delegated
 
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 [![Docker image, latest](https://img.shields.io/docker/image-size/invoiceninja/invoiceninja/latest?label=latest)](https://hub.docker.com/r/invoiceninja/invoiceninja)
 [![Docker image, alpine](https://img.shields.io/docker/image-size/invoiceninja/invoiceninja/alpine?label=alpine)](https://hub.docker.com/r/invoiceninja/invoiceninja)
 
-# Docker for [Invoice Ninja](https://www.invoiceninja.com/) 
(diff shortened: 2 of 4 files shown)


docker-compose.yml
```diff
@@ -32,24 +32,22 @@ services:
...
     image: mysql:5
+#    When running on ARM64 use MariaDB instead of MySQL
+#    image: mariadb:10.4
 #    For auto DB backups comment out image and use the build block below
 #    build:
 #      context: ./config/mysql
     ports:
       - "3305:3306"
     restart: always
-    environment:
```
lwj5, line 48: Can you retain these variables but change the values to point to env vars like this: 

https://github.com/invoiceninja/dockerfiles/blob/1d53d2cbb205f831d0a22448f94e5bd04ad4d690/docker-compose.yml#L51

Include MARIA_USER and so on uncommented below. That way there’s no need to worry about switching between MARIA and MYSQL env.

For the ROOT_PASSWORD can you reference ${DB_ROOT_PASSWORD} instead

lwj5 (commented):
Let me know if I’m unclear.

I can’t comment on the dockerfile maybe @turbo124 can help.

Finally, the error you are experiencing is normal. Because the DB starts later than the IN container. I wanted to resolve by killing the container of migration fails but `php artisan migrate —force` always returns code 0... so if anyone has a solution let me know.

turbo124:
DB failure on first start is normal. Otherwise looks good

docker-compose.yml
tacticz: @lwj5, thanks for the remark, but I'm unsure I understand you well 😕 

If I get you well, your are talking about including (uncommented) `MARIA_***` values in the `docker-compose.yml` file. Why would that be?\
If one refers to the [dockerhub.com MariaDB page](https://hub.docker.com/_/mariadb), it is clearly indicated under the **Environment Variables** section, that these variables are **_strictly identical_** to the ones used by the [MySQL docker image](https://hub.docker.com/_/mysql), i.e.: the MariaDB image makes use of the exact same environment variables as would the MySQL image, I cannot see the point in adding (unused) `MARIA_***` values.

So when you say:
> That way there’s no need to worry about switching between MARIA and MYSQL env.

It is in fact **already the case** with the proposed configuration, since MariaDB is _willingly trying to be as MySQL compatible as possible_.

Regarding your remark saying:
> Can you retain these variables but change the values to point to env vars

This might indeed bring a bit more clarity to the environment variables declaration, and to be honest I myself found it a bit disturbing declaring some values twice but didn't think of referencing them using `${VAR_NAME}` 😉 

As soon as I get your feedback regarding your proposed `MARIA_***` values, I'll update my PR accordingly...

lwj5: From #307 

> I've checked both MySQL's and MariaDB's dockerhub pages and I can confirm that both Environment Variables and Docker Secrets sections are a copy/paste of each others with only the term mysql replaced by mariadb.

i understood from you that the prefixes has to be replaced. I went to check on docker hub and it seems it’s not the case so all is good.

tacticz: @lwj5, thank you for the merge!
> i understood from you that the prefixes has to be replaced. I went to check on docker hub and it seems it’s not the case so all is good.

Oh, yes sorry, I understand your interpretation of my comment, I'll try to be more explicit next time 😉

PR #325, opened 2021-04-18
Update Readme.md - missing sudo

manakill:
without sudo response is "chmod: changing permissions of 'docker/app/public': Operation not permitted"


lwj5:
Why do you need sudo?

shouldn’t the files and folders be owned by you will you first pull or download them?

manakill:
The Docker directions say:
From the terminal run

chmod 755 docker/app/public
sudo chown -R 1500:1500 docker/app


And with those directions you get the error as stated above. Cannot complete directions otherwise.

lwj5:
Shouldn’t the files and folders be owned by you will you first pull or download them? You don’t need sudo to chmod files you own

manakill:
These questions you are asking, could be answered by looking at the Docker directions, or running real testing on physical R Pi 4 ARM64 like I have done. 

Directions Say:
git clone https://github.com/invoiceninja/dockerfiles.git
cd dockerfiles
then edit env and docker-compose.yml
then sudo commands

Raspian(default Pi OS) has the default user with no privileges - hence sudo is required

manakill:
A separate Raspberry Pi 4 (ARMv7 and ARM64) installation instructions could be helpful than to mix the directions in together.

lwj5:
@tacticz would you be able to comment on this? Did you face the same issue

PR #345, opened 2021-05-12
Automatically create user and skip setup page

lwj5:
**Description of the change**

IN app should not be writing to `.env` as

> the file is not shared between multiple instances/replicas

If the SetupController is ran when multiple replicas are running, the config is only written to one container. 

> the file is not persisted and will be removed upon restarting the container

If the SetupController is ran and, for example, Mailer info is not defined in `env` and only provided in the setup page, the config will be lost when the container is restarted.

All in all, the `env` file in docker-compose should be the source of truth and there should not be an `.env` file in the container changing that.

**Benefits**

1. The env provided to docker-compose/helm chart which is passed as environment variables to the container is the single source of truth. No other files will overwrite these values.
2. The user does not need to re-enter info such as DB, URL, Mailer twice as the definitions in in the `env` file should be used.
3. The user is led directly to the login page, and once the user login, they will be prompted to update their email/first/last name

**Possible drawbacks**

I will need your input and review on this as the `/setup` page will be bypassed and it affects:
1. telemetry
2. terms and conditions checklist 

If you could help me to understand at which step in the user flow does IN collect user telemetry, I can analyse if these changes will cause problems.

As for the T&C, will it be sufficient to print it in the logs? Many applications use the phrase along these lines “by continuing to use this application you are deemed to have agreed to the ...” in the logs.

Additional information

The focus of this PR is on the file `10-init-in.sh`.


turbo124:
@lwj5 

This looks good. We were thinking of doing something similar by hiding the redundant info on the setup page - but i think this is a cleaner way to go. If it is ready to merge, please go ahead and merge. 

Thank you!!

Issue #349, opened 2021-05-15
[Helm] Too many redirects after setting up as HTTPS


ffaen:
**Setup information**

Helm

**Describe the bug**
After an initial setup of invoiceninja using helm, trying to access to login screen through https will fail with `ERR_TOO_MANY_REDIRECTS`

**To reproduce**
1. Install Invoiceninja with `helm install -n invoiceninja invoiceninja invoiceninja/invoiceninja --create-namespace -f values.yaml`
2. Complete setup over http
3. Add HTTPS certificate and `helm upgrade -n invoiceninja invoiceninja invoiceninja/invoiceninja`
4. Navigate to URL

**Expected behavior**
Page should not error, show login screen

**Screenshots/logs**
No screenshots or logs that seem useful, just lots of hits to the login page redirecting to itself

**Docker/Kubernetes/Helm**:
- Output of `helm version`:
```
version.BuildInfo{Version:"v3.5.2", GitCommit:"167aac70832d3a384f65f9745335e9fb40169dc2", GitTreeState:"dirty", GoVersion:"go1.15.7"}
```
- Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.0", GitCommit:"af46c47ce925f4c4ad5cc8d1fca46c7b77d13b38", GitTreeState:"clean", BuildDate:"2020-12-08T17:59:43Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.2", GitCommit:"faecb196815e248d3ecfb03c680a4507229c2a56", GitTreeState:"clean", BuildDate:"2021-01-13T13:20:00Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**
values.yaml:

```yaml
nginx:
  ingress:
    enabled: true
    hostname: ninja.{snip}
    # add in step 3
    tls: true
    annotations:
      cert-manager.io/cluster-issuer: "letsencrypt-prod"
    # end
appKey: {snip}
redis:
  password: {snip}
mariadb:
  auth:
    password: {snip}
    rootPassword: {snip}
global:
  storageClass: csi-cinder-classic
volumePermissions:
  enabled: true
```

lwj5:
Could you set add an extra env

REQUIRE_HTTPS=false

and restart the pod

[...]

PR #351, opened 2021-05-16
Update README to add information on env variables

ffaen:
References #349


lwj5:
Perfect, thanks!! I've also included the fix to your last issue where the customer portal is loading with HTTP.

PR #370, opened 2021-07-06
Automatically set values for Nginx sub-chart based on install name

lwj5:
- Automatically change URL in server block
- Updated dependencies
- Updated k8s versions in tests
- Automatically change staticSitePVC

Pending on:

- https://github.com/bitnami/charts/pull/6846


lwj5:
Issues with CI tests. Tested locally, merge for now

PR #374, opened 2021-07-09
Extend liveness probe delay to prevent first migration failure

lwj5:
Initial migration taking at least 1.5min on some computers
Extend time to prevent pod from being killed during this time.


lwj5:
Chart failing due to external issues

Issue #391, opened 2021-08-21
pod/invoiceninja mysqladmin ping timeout at 12 hours


schroedt:
Hi there,

I noticed a hanging init container of `pod/invoiceninja`. That's due to the default of `connection-timeout=43200` (= 12 hours). As the connection-error would be returned after 12 hours of waiting, I do recommend adding a parameter, e.g. `mysqladmin status -h $DB_HOST -P $DB_PORT **--connect-timeout=10**`:

```
kubectl -n invoice-ninja exec -ti pod/invoiceninja-76d84ff655-6jd42 -c wait-db -- sh
~ $ mysqladmin status -h $DB_HOST -P $DB_PORT **--connect-timeout=5**

mysqladmin: connect to server at 'mariadb.db-maria.svc.cluster.local' failed
error: 'Lost connection to MySQL server at 'handshake: reading initial communication packet', system error: 110'
~ $
```

I am using latest Helm chart for Kubernetes:

```
helm repo add invoiceninja https://invoiceninja.github.io/dockerfiles
helm repo update
helm install invoiceninja invoiceninja/invoiceninja --namespace invoice-ninja \
                --set appKey=***\
                --set redis.auth.password=*** \
                --set userEmail=mail@my-example.de \
                --set userPassword=*** \
                --set mariadb.enabled=false \
                --set externalDatabase.host=*** \
                --set externalDatabase.user=*** \
                --set externalDatabase.password=*** \
                --set externalDatabase.database=*** \
                --set externalDatabase.port=***
```

schroedt:
One more finding: I think current implementation cannot run with `istio`, having mTLS `strict`(otherwise: disable TLS in total » bad choise):

- An init countainer is used » before pod-start, before envoy is deployed
- `wait-db` tries connecting to `mariadb` on a istio-enabled pod
- `mariadb` runs within mTLS (strict) because of `server first`protocol
- wait-db has no up and running istio container and will fail 100% because of missing mTLS via envoy.

schroedt:
Login-Page not compatible with `KeePass` and other extensions:

- `User` + `Password` on first page (or first + second)
- Normally `TOTP` is requested on a dedicated second page (if TOTP is configured, otherwise skipped for not irritating users)
- `TOTP` field has a strange name
- `Secret`

lwj5:
Thanks for the suggestion. I agree with the addition of timeout.

For istio I would suggest to add a flag to disable the init container. 

[...]

PR #392, opened 2021-08-21
 #391 mysqladmin ping timeout from 12 hours to 15 seconds

schroedt:
As discussed at  #391, reduced default timeout (12 h) to 15 seconds for wait-db init container. This can be triggered by istio mTLS strict configuration, which requires an up and running envoy proxy. Envoy is started after init-phase.


schroedt:
Hi there,

hope I did it correctly! My first two pull-requests. :-)

lwj5:
Yup! All good to go

PR #396, opened 2021-09-17
Correct timezone in container needed, added tzdata package to respect timezone via `TZ` Env Variable

alexanderhofstaetter:


alexanderhofstaetter:
#397

turbo124:
The app does all its calculations in UTC +0, changing to a different timezone may impact a lot of calculations inside the app.

turbo124:
@lwj5 do you agree we should keep the container @ UTC?

lwj5:
@alexanderhofstaetter is there any specific use case, that you require setting timezone in the docker container?

PR #409, opened 2021-11-06
Default Caddy close #408

### LINKED ISSUE #408: Where to get started with Caddy?
The documentation and every part of this Repository tell me set the default webserver would be caddy but only this line of the config looks like caddy: 

https://github.com/invoiceninja/dockerfiles/blob/c17e37c4b8856066c68de562a8791c99df05f84f/docker-compose.yml#L10

if i change the webserver to caddy and uncomment this line, the caddy is running but not loading the Caddyfile becouse it is the wrong mounting place.

where to find the "default" caddy config template?

kreativmonkey:
Change the docker-compose template back to caddy as the default ``server``.

Every part of the Dockerfiles repository pointed out that caddy should be the default setting but the docker-compose file use nginx as default. This is a working setup with caddy as the default webserver.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,13 +2,13 @@ version: '3.7'
 
 services:
   server:
-    image: nginx
+    image: caddy
     restart: always
     env_file: env
     volumes:
       # Vhost configuration
-      #- ./config/caddy/Caddyfile:/etc/caddy/Caddyfiledocker-com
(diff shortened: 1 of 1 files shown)


turbo124:
Most users are preferring the nginx as the base, so I don't think we'll merge this unfortunately.

kreativmonkey:
In this case you should please update your README and other documentation to fit to the new orientation! Everything tells "default Caddy" and you don't want default caddy, that is not the best experience!

Issue #452, opened 2022-07-20
Error building the mysql backup build


Psycho0verload:
When you try to create the build for the MYSQL backups via docker compose build you get an error:
```
docker /bin/sh: apt-get: command not found
```

The error is due to the image used. The Mysql:5 uses Oraclelinux which works with yum and not with apt

Psycho0verload:
Fixed with Merge #453

Psycho0verload:
With the merge #453 the problem was solved. With the commit [4b82f75](https://github.com/invoiceninja/dockerfiles/commit/4b82f750fcf86012c6125172fcc0a154347d1da6) this problem now exists again and the error as described above occurs again.

```
 ⠿ db Pulled                                                                                                                                                                                                     8.6s
   ⠿ 9815334b7810 Pull complete                                                                                                                                                                                  2.7s
   ⠿ f85cb6fccbfd Pull complete                                                                                                                                                                                  2.7s
   ⠿ b63612353671 Pull complete                                                                                                                                                                                  2.8s
   ⠿ 447901201612 Pull complete                                                                                                                                                                                  3.0s
   ⠿ 9b6bc806cc29 Pull complete                                                                                                                                                                                  3.1s
   ⠿ 24ec1f4b3b0d Pull complete                                                                                                                                                                                  3.1s
   ⠿ 207ed1eb2fd4 Pull complete                                                                                                                                                                                  3.7s
   ⠿ 27cbde3edd97 Pull complete                                                                                                                                                                                  3.7s
   ⠿ 0a5aa35cc154 Pull complete                                                                                                                                                                                  6.2s
   ⠿ e6c92bf6471b Pull complete                                                                                                                                                                                  6.3s
   ⠿ 07b80de0d1af Pull complete                                                                                                                                                                                  6.3s
[+] Building 1.0s (5/6)
 => [internal] load build definition from Dockerfile                                                                                                                                                             0.0s
 => => transferring dockerfile: 324B                                                                                                                                                                             0.0s
 => [internal] load .dockerignore                                                                                                                                                                                0.0s
 => => transferring context: 2B                                                                                                                                                                                  0.0s
 => [internal] load metadata for docker.io/library/mysql:5                                                                                                                                                       0.0s
 => [1/3] FROM docker.io/library/mysql:5                                                                                                                                                                         0.1s
 => ERROR [2/3] RUN apt-get update;                                                                                                                                                                              0.8s
------
 > [2/3] RUN apt-get update;:
[...]

PR #453, opened 2022-07-20
[BUGFIX] Error building the mysql backup build

Psycho0verload:
Bugfix for #452 
Change MYSQL5 to MYSQL5-Debian so that build works for the backup cronjob.
I have implemented this in a running system and the migration worked without problems


Psycho0verload:
Add Fix for #454

turbo124:
Thanks!

almereyda:
For the fix for #454, I'd suggest not to use quotes around the asterisk, if not necessary. Docker Compose will interpret them literally, when read from an `env` file. Also see https://docs.docker.com/compose/compose-file/#env_file-format

maejok-xx:
Was there ever a solution for this? It's still an issue almost a year later.

It looks like from the previous comments the `mysql:8-debian` image is not compatible...is that correct?

PR #455, opened 2022-07-22
[FEATURE] Add support for mysql version 8

Psycho0verload:
I thought it might be handy to be able to use mysql 8 for further developments.

I have switched from mysql 5 to 8. Consciously to version 8.0 because in future versions the support of the small detour via 
```
command: --default-authentication-plugin=mysql_native_password
```
is omitted.

### DIFF
--- a/config/mysql/Dockerfile
+++ b/config/mysql/Dockerfile
@@ -1,4 +1,4 @@
-FROM mysql:5-debian
+FROM mysql:8.0-debian
 # When running on ARM64 use MariaDB instead of MySQL
 #FROM mariadb:10.4
 ENV force_color_prompt yes
--- a/config/mysql/my.cnf
+++ b/config/mysql/my.cnf
@@ -1,0 +1,3 @@
+[mysqld]
+default-authentication-plugin=mysql_native_password
+bind-address=0.0.0.0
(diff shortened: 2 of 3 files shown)


Flightkick:
Instead of downgrading to `mysql_native_password`, support for `caching_sha2_password` may be added by installing the `mariadb-connector-c-dev` apk package here:

https://github.com/invoiceninja/dockerfiles/blob/4b82f750fcf86012c6125172fcc0a154347d1da6/alpine/5/Dockerfile#L53-L60

Psycho0verload:
> Instead of downgrading to `mysql_native_password`, support for `caching_sha2_password` may be added by installing the `mariadb-connector-c-dev` apk package here:
> 
> https://github.com/invoiceninja/dockerfiles/blob/4b82f750fcf86012c6125172fcc0a154347d1da6/alpine/5/Dockerfile#L53-L60

> Instead of downgrading to `mysql_native_password`, support for `caching_sha2_password` may be added by installing the `mariadb-connector-c-dev` apk package here:
> 
> https://github.com/invoiceninja/dockerfiles/blob/4b82f750fcf86012c6125172fcc0a154347d1da6/alpine/5/Dockerfile#L53-L60

Yes you are right about that!

PR #462, opened 2022-10-02
[FEATURE] Add native support MYSQL8

Psycho0verload:
After a feedback from [Flightkick](https://github.com/Flightkick) in #455 I have looked into the native support of MYSQL8 again and tested it successful.

Tested:
- Still works with MYSQL5
- Update from MYSQL5 to MYSQL8 worked without errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ All that is left to do now is bring up the container
 
 ### Running on ARM64 (Raspberry Pi 4)
 
-When deploying on an ARM64 system, you need to comment out the `image: mysql:5` line and uncomment `image: mariadb:10.4` in the `docker-compose.yml` file.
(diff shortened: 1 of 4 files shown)


turbo124:
@lwj5 

Does this look ok?

lwj5:
@Psycho0verload do you know what’s the error message when you encountered this? And which version of IN were you using?

It seems that from php 7.2.4 onwards, caching sha2 has been added to mysqlnd. Right now the containers are using php 8.1

In any case I’ll try it out tmr.

Psycho0verload:
I tested it and interestingly I got the error that the sha2 module is missing. I do not have the exact error message at hand because I am not sitting at the notebook.

It printed in the error message that it would like to have the sha2 module in the path of mariadb. We get the needed module with the installed package.

lwj5:
Tried it out and works, this is required even with later PHP versions

PR #472, opened 2022-11-18
Bump helm version to 5.5

xeruf:


lwj5:
You would need to modify the chart version as well.

See https://github.com/invoiceninja/dockerfiles/pull/437/files

xeruf:
Ah makes sense. I think it's fine to omit the minor version - people can still pin it themselves, and this way we don't have to update it as frequently.

lwj5:
I'll close this PR since there's no update.

PR #517, opened 2023-06-23
Please don't bind mysql to all interfaces with default credentials in Docker-compose

Leopere:
Please don't do this. Default credentials on the sql container in the .env file and then bind blindly 3305 on all interfaces for the public  to log in with ninja//ninja

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -44,8 +44,6 @@ services:
 #    For auto DB backups comment out image and use the build block below
 #    build:
 #      context: ./config/mysql
-    ports:
-      - "3305:3306"
     restart: always
     env_file: env
     volumes:


ryanwinter:
What's holding this back from merge? I noticed this immediately when reviewing the docker compose.

Leopere:
@ryanwinter 
I can't begin to imagine why you would want to expose your potential user base to such a blatant security misconfiguration that is entirely unnecessary it honestly makes me question how safe this software even is.  

This is by far the best software for this task that I can find but woooof that's roouuugh.

turbo124:
Thanks!

Leopere:
y'all rock tyvm <3 I very much appreciate this tool its absolutely the best in class.

PR #519, opened 2023-07-06
Adds options to use config bases php.ini

turbo124:
Allows users to easily configure php  from config file.


lwj5:
alpine/5/cache_buster

Maybe you can try putting a new datetime into the file to get the latest php stuff

turbo124:
@lwj5 

builds are falling over due to mysql on arm issues i think

lwj5:
Error is on AMD64

The error is when composer install is completed it runs Artisan discover. Not sure why artisan discover tries to connect to DB. There's no DB in the container, and it shouldn't be connecting to a DB at that stage

turbo124:
@lwj5 

I'm baffled by this error, the only workaround I can currently engineer is to disable the scripts for composer, and then perform the package discovery in the init.

There have been no significant changes to the composer.json file, and also using composer update in the build still generates the same issue. Any other thoughts?

PR #551, opened 2023-12-22
feat(helm): permit specifying additionalServerBlockConfig

clementnuss:
permits configuring an additional serverBlockConfig for nginx which can be used to prevent caching of PDFs when using a CDN (such as Cloudlfare) in front of InvoiceNinja, as documented in the troubleshooting section:

https://invoiceninja.github.io/en/self-host-troubleshooting/#pdfs-dont-appear-to-be-updating


turbo124:
@clementnuss in the most recent version, we no longer cache the PDFs, they are generated dynamically now so this should not be required

clementnuss:
ah great! then there's no point, indeed 👍🏼

PR #563, opened 2024-02-04
Specified compatible nginx version

DinkDonk:
The nginx configuration used by Invoice Ninja is not compatible with the latest nginx Docker image. Changing docker-compose.yml to lock nginx to v1.25 which is known to work.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3.7'
 
 services:
   server:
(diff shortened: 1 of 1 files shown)


turbo124:
@DinkDonk 

What error are you seeing? I note the latest nginx on docker is 1.25.3 so I would assume locking to 1.25 should be essentially the same?

PR #567, opened 2024-03-22
Bump helm version to 5.8.37

sq7lrx:


sq7lrx:
I've checked this Helm chart on Kubernetes 1.28. Both fresh install and upgrade from chart 0.10.2 work OK

PR #572, opened 2024-04-04
Updates for react release

turbo124:


BloodyIron:
Hey so I'm trying to figure out... but did this commit enable the "Web App" (React) button along the top I see now? I'm getting an error trying to use that and I'm trying to figure out how to switch to the react app (running the docker image in k8s). It seems to do an HTTP PUT that fails when clicking on the button, and the button showed up when I grabbed an updated image from dockerhub, but I don't remember what version I was on before (months old probably). I can't find any documentation on the React web app aspect so reverse-engineering the situation.

PR #579, opened 2024-05-28
Update README.md

Zuhdan:
add another inline command to configure the local storage


turbo124:
i don't think is needed at all as long as the base folder permissions are set correctly.

Zuhdan:
Yeah i think so.
Thanks

PR #594, opened 2024-06-24
Simplified Chinese Fonts Added

Kingwayer:
Resubmit PR, please review and approve.

### DIFF
--- a/alpine/5/Dockerfile
+++ b/alpine/5/Dockerfile
@@ -71,7 +71,7 @@ CMD ["/bin/sh"]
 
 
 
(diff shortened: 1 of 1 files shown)


alpine/5/Dockerfile
```diff
@@ -44,9 +48,20 @@ RUN set -eux; \
     font-isas-misc \
     supervisor \
     mysql-client \
-    chromium \
```
turbo124, line 47: why was chromium removed?

turbo124 (commented):
why was chromium removed?

PR #656, opened 2024-11-26
Use php.ini-production

benbrummer:
At the moment no php.ini is located at /usr/local/etc/php/php.ini. Based on the documentation https://hub.docker.com/_/php
image
php.ini-production will be used now

### DIFF
--- a/debian/Dockerfile
+++ b/debian/Dockerfile
@@ -3,17 +3,16 @@ ARG saxon=12.5.0
 
 # Install system dependencies
 RUN apt-get update && apt-get install -y \
-    git \
     curl \
-    libpng-dev \
-    libonig-dev \
-    libxml2-dev \
-    zip \
-    unzip \
-    gosu \
     default-mysql-client \
-    supervisor \
     fonts-liberation \
+    fonts-noto-cjk \
+    fonts-noto-cjk-extra \
+    fonts-wqy-microhei \
+    fonts-wqy-zenhei \
+    git \
+    gnupg2 \
+    gosu \
     libasound2 \
     libatk-bridge2.0-0 \
     libatk1.0-0 \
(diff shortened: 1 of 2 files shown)


benbrummer:
https://github.com/invoiceninja/dockerfiles/pull/656/commits/24d49a4374a16ddaf6137bfc23002124df5b2454:
- Sort apt packages and php modules alphabetical
- Do not install php modules, which are already installed in the base image
- Install install-php-extensions from github
- Clean up temporary files from saxon installation
- use --chmod=0755 in COPY and ADD

https://github.com/invoiceninja/dockerfiles/pull/656/commits/535d6d4c06f4a67a1babd29fe8b9659737e4b36e
- remove deprecated version
- remove php.ini mounting again
- rename image-public to image_public to align volume names

benbrummer:
https://github.com/invoiceninja/dockerfiles/pull/656/commits/6274c432d9ae03eef920f94a131831f76d4c3e43:
- fix nginx symlink https://github.com/invoiceninja/dockerfiles/commit/6274c432d9ae03eef920f94a131831f76d4c3e43

turbo124:
@benbrummer In the last PR, i did change to mounting the php.ini we provide here:

```
      - ./php/php.ini:/usr/local/etc/php/php.ini
```

Would there be any reason to change?

benbrummer:
> @benbrummer In the last PR, i did change to mounting the php.ini we provide here:
> 
> ```
>       - ./php/php.ini:/usr/local/etc/php/php.ini
> ```
> 
> Would there be any reason to change?

https://github.com/invoiceninja/dockerfiles/blob/60fd0aa79ebc7be34d4f286f2c49f9e09d0a3648/debian/Dockerfile#L92 copies your file already into the image. With the mount in docker-compose.yaml. PDF was not working anywhere for me, no preview + no attachment.

Following the official documentation we should start with php.ini-production
image

and mount `- ./php/php.ini:/usr/local/etc/php/conf.d/app.ini` we could remove the app.ini from the Dockerfile (or leave it as fallback). The mount is good, as we can change php configuration without rebuilding the entire image.

turbo124:
Thanks!

PR #658, opened 2024-11-26
move client_max_body_size from default.conf to nginx.conf

brdns:
- [Module ngx_http_core_module](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size)
- add `client_body_buffer_size`
- prevents the following logs for each upload : `[warn] a client request body is buffered to a temporary file`

### DIFF
--- a/debian/docker-compose.yml
+++ b/debian/docker-compose.yml
@@ -31,7 +31,6 @@ services:
       resources:
         limits:
           memory: 512M
-          #memory: 1G # prevents swapping on pdf generation with chrome
     logging: *default-logging
 
   nginx:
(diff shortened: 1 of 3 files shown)


turbo124:
Thanks!

benbrummer:
@brdns may I ask how big are the files, which you upload?

I think about a sufficient configuation for invoiceninja should be 
### nginx
client_max_body_size = 20
### php-fpm
client_body_buffer_size = 20M
post_max_size = 20M
upload_max_filesize = 20M => post_max_size will be the limit, as the header-size is included

benbrummer:
@brdns 10M should be a good default. Do you have some recommendation regarding php memory_limit. I have no issues with the default 128M.

PR #659, opened 2024-11-27
install saxon 12.5.0 with install-php-extensions

benbrummer:
- install saxon with install-php-extensions
- replace deprecated apt-key
- artisan optimize will handle caches https://laravel.com/docs/11.x/deployment#optimization

### DIFF
--- a/debian/Dockerfile
+++ b/debian/Dockerfile
@@ -40,8 +40,9 @@ RUN apt-get update && apt-get install -y \
     xfonts-wqy \
     zip \
     && if [ "$(dpkg --print-architecture)" = "amd64" ]; then \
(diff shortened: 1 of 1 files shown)


debian/Dockerfile
```diff
@@ -73,19 +74,18 @@ ADD --chmod=0755 https://github.com/mlocati/docker-php-extension-installer/relea
...
     gmp \
     imagick \
-    # mbstring \ Already installed
     mysqli \     
     opcache \
     pcntl \
     pdo_mysql \
     redis \
-    # saxon \ outdated
+    saxon-${saxon} \
```
turbo124, line 88: Does this really work?

I was under the impression that you would still need libsaxon in the system also? Or does this extension also pull this in?

I'd be pretty happy if it did!

turbo124:
Thanks!

PR #661, opened 2024-11-28
php 8.3 and apt package optimization

benbrummer:
- With this pr php will be updated to 8.3
- memory limits for containers will be removed, as these conflict with php/opcache configuration and cause random 500
- ap packages for amd64 have been reduced to thge required ones
- build context has been added to docker compose file, to make it easily deployable with a local build
- Fix for `WARN For [program:scheduler], redirect_stderr=true but stderr_logfile has also been set to a filename, the filename has been ignored`

Unchanged behaviour, use latest build from docker hub.
```
docker compose up
```

Local build with latest tag
```
docker compose up --build will build the image with latest tag
```

Local build with custom tag
```
export TAG=some_value
docker compose up --build
```

Get updated base images for the build
```
docker compose up --pull --build
```


hillelcoren:
Hey, just wanted to thank you for all of your contributions!

turbo124:
@benbrummer looks good, i think we may have lost the soap extension along the way, this is a nice to have for vat number verifications.

turbo124:
@benbrummer 

We might have to roll this back I am afraid. When i build the image from scratch, i lose access to the mysql client at the command line which is required when running the initial script to write the schema and seed the database.

PR #663, opened 2024-11-29
Corrective fixes for new dockerfile

turbo124:
Fixes for nginx response buffers, and remove if checks around architecture that prevented mysql client from installing.

### DIFF
--- a/debian/Dockerfile
+++ b/debian/Dockerfile
@@ -8,44 +8,40 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     gnupg2 \
     gosu \
     supervisor \
+    default-mysql-client \
+    fonts-liberation \
+    fonts-noto-cjk \
+    fonts-noto-cjk-extra \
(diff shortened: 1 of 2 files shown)


benbrummer:
Response buffers are now handled twice

https://github.com/invoiceninja/dockerfiles/blob/113c700754e5db0c95d8e3fc6738c75c7fd6fc11/debian/nginx/invoiceninja.conf#L4
and
https://github.com/invoiceninja/dockerfiles/blob/113c700754e5db0c95d8e3fc6738c75c7fd6fc11/debian/nginx/laravel.conf#L26

Cuase of the alphabetical order laravel.conf will be the effective ones

turbo124:
I don't think those were loading in the correct block. The error message persisted until I added the reference to the fastcgi block in the vhost itself

benbrummer:
@turbo124 for me the fastcgi_buffer_size is still to small. Same behavior with setting it at either location. But it happens only with the desktop app.

Issue #660, opened 2024-11-27
Dockerfile optimization proposal


benbrummer:
docker image seems to be quite big, so I try to identify ways to shrink it

### Ideas
- apt-get --no-install-recommends
- Review apt packages: 
  - snappdf requirements: https://github.com/beganovich/snappdf?tab=readme-ov-file#headless-chrome-doesnt-launch-on-unix
- Review php modules
  - https://invoiceninja.github.io/en/self-host-installation/
- https://docs.docker.com/build/building/multi-stage/
- php 8.3

### apt packages
required:
- curl => google-chrome-stable
- gnupg2 => google-chrome-stable
- gosu => init.sh
- default-mysql-client => initial user creation

obsolete (?):
- git
- unzip
- wget
- zip
- ...

### php modules
```
[PHP Modules]
bcmath
Core
ctype
curl
date
dom
exif
fileinfo
filter
gd
gmp
hash
iconv
imagick
json
libxml
mbstring
mysqli
mysqlnd
openssl
pcntl
pcre
PDO
pdo_mysql
pdo_sqlite
Phar
posix
random
readline
redis
Reflection
saxonc
session
SimpleXML
soap
sodium
SPL
sqlite3
standard
tokenizer
xml
xmlreader
xmlwriter
Zend OPcache
zip
zlib
[Zend Modules]
Zend OPcache
```

### Misc
- https://github.com/invoiceninja/dockerfiles/blob/a6d8f663ae24b4ed982c68e68deb258c55ab6863/debian/Dockerfile#L149 seems to do nothing
- Exact versions for php modlues e.g saxon vs no version for other modules?

turbo124:
@benbrummer we can certainly start with the obsolete ones like zip/unzip

turbo124:
@benbrummer 

related #663 I had to roll back the architecture if/fi blocks as they were preventing required dependencies from installing.

turbo124:
The fonts are definitely needed to provide unicode support for PDF generation is regions like China.

The others, if they are dependencies of chrome won't hurt to have in place I would have thought.

benbrummer:
@turbo124

Are these apt packages needed?
I assume the *-dev can be ommited and xfonts-wgy is propably required for pdf?
```
    libonig-dev \
    libpng-dev \
    libxml2-dev \
    xfonts-wqy \
```

turbo124:
I'm not sure if any of these are subdependencies of anything

[...]

PR #664, opened 2024-11-29
move google-chrome-stable related packages to arm64 section

benbrummer:
https://github.com/invoiceninja/dockerfiles/issues/660#issuecomment-2507312764

Remove unnecessary `php artisan config:cache` from `init.sh`, as it is handled with `php artisan optimize`

### DIFF
--- a/debian/Dockerfile
+++ b/debian/Dockerfile
@@ -5,36 +5,17 @@ ARG saxon=12.5.0
 # Install system dependencies
 RUN apt-get update && apt-get install -y --no-install-recommends \
     curl \
+    default-mysql-client \
     gnupg2 \
     gosu \
     supervisor \
-    default-mysql-client \
-    fonts-liberation \
     fonts-noto-cjk \
     fonts-noto-cjk-extra \
     fonts-wqy-microhei \
     fonts-wqy-zenhei \
-    libasound2 \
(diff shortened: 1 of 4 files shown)


turbo124:
@benbrummer for me, when i just pulled in the latest image, the newest react files did not  get brought over.

This is because we are not using a volume pointing at public/ even thou it is read only, this excludes the image public directory from being mounted.

benbrummer:
> @benbrummer for me, when i just pulled in the latest image, the newest react files did not get brought over.
> 
> This is because we are not using a volume pointing at public/ even thou it is read only, this excludes the image public directory from being mounted.

So the react files should be updated in the volume image_public:/var/www/html/public:ro?

turbo124:
@benbrummer 

I've created a PR with a  new approach here:

#665

this will copy the /public directory of the image to a temp dir, then when the entrypoint script it reach, we'll remove the current js/css files from public/ and copy the contents of the public/ directory over to the volume.

thoughts on this approach? 

This is the only way I can think to work around the volume mount being _required_ for the public/ directory and also allowing the newer files in the image to be available.

benbrummer:
Fix for https://github.com/invoiceninja/dockerfiles/pull/664/

turbo124:
Thanks for your help on this!

PR #665, opened 2024-11-30
Update dockerfile

turbo124:


benbrummer:
I think we have have only the option to have a single nginx-php-fpm conatiner and only mount the storage volume
or have two containers mounting the appölication from https://github.com/invoiceninja/invoiceninja, after it has been added as submodule to this repo.

```
# Ensure symlink for storage/app/public
if [ ! -L /var/www/html/public/storage ]; then
  echo "Creating symlink for storage/app/public..."
  ln -sfn /var/www/html/storage/app/public /var/www/html/public/storage
fi
```
could be replaced with  `php artisan storage:link`

If we go with  a single container, we could go with frankenphp.

turbo124:
@benbrummer I wanted to avoid comingling nginx + php images as we've had criticism on that front before as it then is not optimal for a user who want to use caddy etc etc. 

The submodule idea is good, but we have the added complexity here of merging the react project also. This would not be a deal breaker, but it does add some additional overhead I think. I've never tinkered with this kind of idea. However I think ultimately we are still in trouble as we will always need to expose the public/ directory even if :ro which then circles us back to the original problem where the volume mount excludes the images public/ dir....

benbrummer:
With the submodule, we only need to update the commit hash. To get the new version.

benbrummer:
> I wanted to avoid comingling nginx + php images as we've had criticism on that front before as it then is not optimal for a user who want to use caddy etc etc.

1. Base image
invoiceninja/base
2. extend the base
from invoiceninja/base => invoiceninja/nginx, incoiceninja/apache, incoiceninja/caddy...

benbrummer:
I will try to come up with a proposal.

Can you tell me the purpuose of `/var/www/html/public/uploads`, https://laravel.com/docs/11.x/structure. As it is not in the storage or public/storage folder it will not persist. It is created at https://github.com/invoiceninja/dockerfiles/blob/113c700754e5db0c95d8e3fc6738c75c7fd6fc11/debian/Dockerfile#L127

turbo124:
> I will try to come up with a proposal.
> 
> Can you tell me the purpuose of `/var/www/html/public/uploads`, https://laravel.com/docs/11.x/structure. As it is not in the storage or public/storage folder it will not persist. It is created at
> 
> https://github.com/invoiceninja/dockerfiles/blob/113c700754e5db0c95d8e3fc6738c75c7fd6fc11/debian/Dockerfile#L127
> 
> as it is not inside `/var/www/html/public/storage`, this folder is not persitently stored. Is it just a temporary upload location?

I believe public/uploads it is redundant.

I will merge this PR as a temporary solution so that as we tag new images, the files can at least update until a new master image has been created.

PR #676, opened 2024-12-21
Make arm64 available for usage

benbrummer:


benbrummer:
@turbo124 I think it would make sense to use chromium for arm64 and amd64. As the debian maintained package is installed it is a stable package.

turbo124:
Thanks!

PR #678, opened 2024-12-28
Frankenphp

benbrummer:
- frankenphp
  - Avoids workaround for react updates
  - Worker mode for Laravel Octane (invoiceninja needs to be made ready for it)
  - gzip, brotli and zstd for static files
 - mariadb 11.4 (lts)
 - valkey 8 (redis)
 - Removes supervisord
   - Use separate containers for worker/scheduler   
image


turbo124:
@benbrummer 

This looks awesome!

My vote for this one would be to create a new branch and build this as an optional image. There will need to be some careful testing around this due to the long lived PHP process approach with octane.

benbrummer:
@turbo124 Do you have a suggestions for the branch-name?

Locally I'm already able to run with Laravel Octane, but I have an issue with
`composer require laravel/octane` if I do not set a my GitHub token in advance with `composer config --global github-oauth.github.com TOKEN`.

image

With --prefer-source
image

turbo124:
Let's use octane for the branch name, I will include Laravel octane in the dependencies

benbrummer:
- composer is now handled in a separate stage
- octane was tested locally but fails at the moment to install without providing a github token

debian/docker-compose.yml
```diff
@@ -1,47 +1,78 @@
...
 x-logging: &default-logging
   options:
     max-size: "10m"
     max-file: "3"
   driver: json-file
 
+x-app-volumes: &volumes
+  volumes:
+      - ./.env:/app/.env
+      - ./php/php.ini:/usr/local/etc/php/conf.d/zzz-php.ini:ro
```
stavros-k, line 12: Do you think having a default zzz-1-php.ini baked in the image would be better?
And users could mount additional ini's to set more or override things like zzz-2-php.ini

turbo124:
@benbrummer 

The next time we tag a release, the octane dependency will be automatically handled from the main invoiceninja/invoiceninja repo.

debian/docker-compose.yml
benbrummer: Makes totaly sense to delivere a base php-file inside the image. Will be integrated with the next PR. Will do it as `zzz-0-php.ini`.

PR #682, opened 2025-01-04
Cleanup

benbrummer:
- octane has been integrated in composer.json, so no need to install it in dockerfile
- php.ini cleanup

### DIFF
--- a/debian/Dockerfile
+++ b/debian/Dockerfile
@@ -1,4 +1,4 @@
-FROM composer:latest AS composer
+FROM dunglas/frankenphp:1-php8.3-bookworm AS prepare-app
 
 RUN curl -s "https://api.github.com/repos/invoiceninja/invoiceninja/releases/latest" | \
     grep -o '"browser_download_url": "[^"]*invoiceninja.tar"' | \
@@ -10,11 +10,6 @@ RUN ln -s ./resources/views/react/index.blade.php ./public/index.html
 
 # Set permissions: directories 755, files 644
 RUN chmod -R a=r,u+w,a+X .
-
-# Install dependencies
-RUN composer install --no-dev --no-scripts --no-autoloader --ignore-platform-reqs
-
-RUN composer dump-autoload --optimize
 
 RUN php artisan storage:link
 
@@ -93,7 +88,7 @@ ENTRYPOINT ["/usr/local/bin/init.sh"]
 CMD ["frankenphp", "php-cli", "artisan", "octane:frankenphp"]
 
 # InvoiceNinja
-COPY --from=composer --chown=${user}:${user} /app /app
+COPY --from=prepare-app --chown=${user}:${user} /app /app
 
 # Add initialization script
 COPY --chmod=0755 scripts/init.sh /usr/local/bin/init.sh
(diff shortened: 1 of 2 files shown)


benbrummer:
@turbo124 i had a look on https://github.com/invoiceninja/invoiceninja/actions/workflows/react_release.yml to try to get rid of some reduntant things and have some findings/questions

1. Looks like there is no need to run `composer install --no-dev --no-scripts --no-autoloader` as dependencies are already installed in https://github.com/invoiceninja/invoiceninja/actions/runs/12609050122/workflow#L31
1. `composer dump-autoload --optimize` is obsolete as composer.json sets `"optimize-autoloader": true`
=> Composer is not relevant for dockerfiles, so the composer stage can be removed

1. There are two errors in `Prepare Laravel Application` https://github.com/invoiceninja/invoiceninja/actions/runs/12609050122/workflow#L33
    1. https://github.com/invoiceninja/invoiceninja/actions/runs/12609050122/job/35142179418#step:6:27
    1. https://github.com/invoiceninja/invoiceninja/actions/runs/12609050122/job/35142179418#step:6:30
=> I think the whole `Prepare Laravel Application` section is not necessary at all
=> artisan optimze is removed with cache:clear and again with https://github.com/invoiceninja/invoiceninja/actions/runs/12609050122/workflow#L69
=> storage:link is not needed and excluded for the tar https://github.com/invoiceninja/invoiceninja/actions/runs/12609050122/workflow#L78
=> Exclude for .htaccess is wrong, .htaccess is added to invoiceninja.tar https://github.com/invoiceninja/invoiceninja/actions/runs/12609050122/workflow#L78. I think it should not be excluded, anyway.

1. `Cleanup Builds` https://github.com/invoiceninja/invoiceninja/actions/runs/12609050122/workflow#L67
There could be more candidates for deletion e.g. `.github`, `.gitignore`, `.gitattributes`, maybe the `tests` folder and more.

1. `invoiceninja.tar` https://github.com/invoiceninja/invoiceninja/actions/runs/12609050122/workflow#L78
We could have a tar.gz or tar.zst instead

Besides this, it would be good if we could have permissions 644 for files and 755 for directories
so we do not need to do it in the dockerfile https://github.com/invoiceninja/dockerfiles/blob/a71af868d20c8dda9ddb0415fbb36516b41cf47d/debian/Dockerfile#L12
```
find /app -type f -exec chmod 644 {} \;  
find /app -type d -exec chmod 755 {} \;  
```
or with chown alone
`chmod -R a=r,u+w,a+X /app`

Directories are already fine, but ui folder is not covered as this is added afterwards https://github.com/invoiceninja/invoiceninja/actions/runs/12609050122/workflow#L40

So apart of the findings for https://github.com/invoiceninja/invoiceninja/actions/workflows/react_release.yml, I would remove composer and if permissions are already set to 644/755 inside the tar the chown step, which takes a (too) long time.

turbo124:
@benbrummer 

Looks good, i think we'd always want to chown in the docker workflow to ensure isolation between the workflows.

benbrummer:
> @benbrummer
> 
> Looks good, i think we'd always want to chown in the docker workflow to ensure isolation between the workflows.

I agree for chown, which is handled by `COPY --from=prepare-app --chown=${user}:${user} /app /app`, but nor for chmod.

I will make a proposal for https://github.com/invoiceninja/invoiceninja/actions/workflows/react_release.yml

PR #686, opened 2025-01-09
healthcheck for scheduler and worker

benbrummer:


benbrummer:
Firefox HTTP/3 Bug https://bugzilla.mozilla.org/show_bug.cgi?id=1896216

turbo124:
@benbrummer 

was there something to address with the http3 issue prior to merge?

benbrummer:
> @benbrummer
> 
> was there something to address with the http3 issue prior to merge?

No it is not a specific issue with invoiceninja, but a bug with Firefox, which we need to workaround:

Three options
1. No port mapping for 443/udp => works for everyone
2. Firefox users need to disable HTTP/3 in the browser config about:config
3. Firefox users need to switch to a chromium based browser

turbo124:
Thanks!

PR #689, opened 2025-01-13
Align alpine to debian and octane for maintenance

benbrummer:
- Updates to php 8.3
- add redis/valkey (without volume)
- Use invoiceninja.tar.gz
- php configuration inside image
- opcache.preload

This deployment suffers from not updating files in the in the the public folder. Wiping the public folder in the same way as we do it in the debian deployment is not possible as we would delete all storage content, too vs a symlink in debian.

So we need to make a similar approach as we have in debian, delete everything but the storage folder and copy new files afterwards. In the alpine deployment it is no symlink, so we need to exclude it.


benbrummer:
pins alpine to 3.20 (eol 2026-04-01) to solve initialization

turbo124:
@benbrummer 

I'm reluctant to make any changes on this branch, it is essentially deprecated and only receiving support for anything that is broken. A large refactor like this may cause issues with existing installations.

benbrummer:
@turbo124 I understand your concerns, but
- in terms of datastorage and database, no breaking change
- rootfs behaviour has not been changed
- valkey/redis was added without persistent data, so no modification persisten volumes are created
- the php-configuration, which is currently baked into the image is aligned without breaking change
- invoiceninja.tar.gz is required to be able to remove the invoiceninja.tar from the react-release.yaml

Bugfix:
- pinning alpine to 3.20 instead of latest, which changed in december to 3.21, fixes the mariadb initialization

Known issue - not fixed:
- react is not updating properly

In total the only important thing is the bugfix for mariadb, which you could cherry pick, so we can close the pr. I can maintain the alpine branch for myself in my fork, which is fine for me.

turbo124:
Thanks!

PR #710, opened 2025-01-24
 align build and publish images

turbo124:


lakelounge:
I pulled the new version several times (over 5.11.53) but always get a white browser window. Cleared the Browser Cache but with no success.

lakelounge:
Did a clean install, but still no success.
Browser Inspector says: BrowserAuthError: crypto_nonexistent: The crypto object or function is not available.
Any Idea?

Bildschirmfoto 2025-03-15 um 16 57 06

sinjap:
I have the same issue as @lakelounge

PR #717, opened 2025-01-28
use invoiceninja.tar.gz

benbrummer:
Afterwards we can remove https://github.com/invoiceninja/invoiceninja/blob/99d2652f89b64c49be3aefe420500e652e50bbaf/.github/workflows/react_release.yml#L66

### DIFF
--- a/alpine/5/Dockerfile
+++ b/alpine/5/Dockerfile
@@ -8,17 +8,17 @@ FROM --platform=$BUILDPLATFORM node:lts-alpine as nodebuild
 # Download Invoice Ninja
 ARG INVOICENINJA_VERSION
 ARG REPOSITORY=invoiceninja/invoiceninja
-ARG FILENAME=invoiceninja.tar
+ARG FILENAME=invoiceninja.tar.gz
 
 RUN set -eux; apk add curl unzip grep
 
-RUN DOWNLOAD_URL=$(curl -s "https://api.github.com/repos/invoiceninja/invoiceninja/releases/latest" | grep -o '"browser_download_url": "[^"]*invoiceninja.tar"' | cut -d '"' -f 4) && \
+RUN DOWNLOAD_URL=$(curl -s "https://api.github.com/repos/invoiceninja/invoiceninja/releases/latest" | grep -o '"browser_download_url": "[^"]*invoiceninja.tar.gz"' | cut -d '"' -f 4) && \
     curl -LJO "$DOWNLOAD_URL" && \
-    mv invoiceninja.tar /tmp/ninja.tar
+    mv invoiceninja.tar.gz /tmp/ninja.tar.gz
 
 # Extract Invoice Ninja
 RUN mkdir -p /var/www/app \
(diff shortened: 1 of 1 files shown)


alpine/5/Dockerfile
```diff
@@ -8,17 +8,17 @@ FROM --platform=$BUILDPLATFORM node:lts-alpine as nodebuild
...
-RUN DOWNLOAD_URL=$(curl -s "https://api.github.com/repos/invoiceninja/invoiceninja/releases/latest" | grep -o '"browser_download_url": "[^"]*invoiceninja.tar"' | cut -d '"' -f 4) && \
+RUN DOWNLOAD_URL=$(curl -s "https://api.github.com/repos/invoiceninja/invoiceninja/releases/latest" | grep -o '"browser_download_url": "[^"]*invoiceninja.tar.gz"' | cut -d '"' -f 4) && \
     curl -LJO "$DOWNLOAD_URL" && \
-    mv invoiceninja.tar /tmp/ninja.tar
+    mv invoiceninja.tar.gz /tmp/ninja.tar.gz
 
 # Extract Invoice Ninja
 RUN mkdir -p /var/www/app \
-    && tar -xvf /tmp/ninja.tar -C /var/www/app/ \
+    && tar -xvf /tmp/ninja.tar.gz --strip-components=1 -C /var/www/app/ \
```
turbo124, line 21: why are we introducing the strip-components=1 here? this will change the directory levels?

turbo124 (changes requested):
Just need clarification on this item

alpine/5/Dockerfile
benbrummer: - ninja.tar directly contains the gzipped files.
- ninja.tar.gz contains an additional folder with the gzipped files. This is aligned with standard conventions. So if someone extracts it careless ly, it will not by default extract into the current directory, which may contain already files, but in a folder in the current directory. So tar.gz with strip-components is identical to tar without strip-components

turbo124: This impacts the selfupdater which is expecting there to be no nested directory. PharData ->extractTo() cannot handle this particular scenario and would require extracting to a temp dir and then copying file back.

Also, changing the directory structure will impact other third parties that use our release files for their own builds.

benbrummer: The release artifact invoiceninja.tar is at the moment still available and the selfupdater is not used in docker.

Proposal:
We should get rid of invoiceninja.tar:
- Unexpected/non-standard behaviour of extracting by default into the current directory
- wrong extension, as it is a gzip archive

So removing https://github.com/invoiceninja/invoiceninja/blob/99d2652f89b64c49be3aefe420500e652e50bbaf/.github/workflows/react_release.yml#L66 needs than be postponed, until consumers of invoiceninja.tar had some time switch to the new artifact.

benbrummer:
> Just need clarification on this item

Maybe we should than just change the extension to reflect the gzip compression and leave the content as before.

turbo124:
@benbrummer 

yes, that sounds perfect.

PR #730, opened 2025-03-15
use github latest url, tar.gz without top level repository, mbstring

benbrummer:
relates to: https://github.com/invoiceninja/dockerfiles/pull/717
depends on: https://github.com/invoiceninja/invoiceninja/pull/10772


turbo124:
@benbrummer we'd also need to update master branch also, correct?

benbrummer:
> @benbrummer we'd also need to update master branch also, correct?

PR for master is ready for review

PR #736, opened 2025-03-23
google-chrome-stable does not exist, replace with chromium

gesa:
I saw a reference to Google Chrome stable being installed in the dockerfile, but ultimately the only binary that’s showed up in my fresh installs in /usr/bin/chromium.

I was getting 500s when attempting to generate PDFs. After changing this and restarting, I now see PDF previews.

If this PR gets treated as an issue and there’s a way better/different fix, that’s cool too.

### DIFF
--- a/debian/.env
+++ b/debian/.env
@@ -57,6 +57,4 @@ NORDIGEN_SECRET_KEY=
 
 IS_DOCKER=true
 SCOUT_DRIVER=null
-SNAPPDF_CHROMIUM_PATH=/usr/bin/google-chrome-stable
-
-
+SNAPPDF_CHROMIUM_PATH=/usr/bin/chromium


turbo124:
Thanks for this, i believe the issue here is platform related, arm64 platforms will use chromium, x86 => google chrome

I'm not sure how we'll be able to gracefully handle this particular scenario @ runtime.... perhaps update the .env with better docs for this could be the solution...

@benbrummer thoughts?

benbrummer:
In my opinion environment variables, which are mandatory to be aligned to the image and therefore not configurable for the user should be baked into the image.
```
FILESYSTEM_DISK=debian_docker
IS_DOCKER=true
SNAPPDF_CHROMIUM_PATH=/usr/bin/google-chrome-stable
```
If someone wants to customize them, e.g. for the FILESYSTEM_DISK it can be overwritten with the .env file.

For the support questions related to SNAPPDF_CHROMIUM_PATH and amd64/arm64, we have multiple solutions:

1. Go with chromium for both platforms
2. Set it conditionally in the init.sh
3. Leave it as it is, but the user will need to take care about the correct .env setting => We need to support every now and than with issues

turbo124:
@benbrummer setting conditionally in the init.sh does make sense.

We've seen flaky behaviour from chromium, and very stable performance from google-chrome-stable, hence the divergence here...

i'll PR this change into debian.

PR #738, opened 2025-03-27
fix: update init script to chown cache directory

mindovermiles262:
Using provided docker-compose file, IN does not start up due to invalid directory permissions, cycling the error message:

```
Updating public folder...
Public Folder is up to date
In PackageManifest.php line 179:
                                                                             
  The /var/www/html/bootstrap/cache directory must be present and writable. 
```

This PR chown's the bootstrap/cache directory that's mounted into the container fixing this issue similar to all the other mounted volumes.

### DIFF
--- a/debian/scripts/init.sh
+++ b/debian/scripts/init.sh
@@ -17,12 +17,15 @@ if [ "$*" = 'supervisord -c /etc/supervisor/conf.d/supervisord.conf' ]; then
     # Ensure owner, file and directory permissions are correct
     chown -R www-data:www-data \
         /var/www/html/public \
-        /var/www/html/storage
+        /var/www/html/storage \
+        /var/www/html/bootstrap/cache
     find /var/www/html/public \
         /var/www/html/storage \
+        /var/www/html/bootstrap/cache \
         -type f -exec chmod 644 {} \;
     find /var/www/html/public \
         /var/www/html/storage \
+        /var/www/html/bootstrap/cache \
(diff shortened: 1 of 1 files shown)


benbrummer (approved):

benbrummer:
I'm not sure if it really makes sense to have bootstrap as volume at all, as https://github.com/invoiceninja/dockerfiles/blob/39c546e270f58e264ea6c5acdd7e084a1771908f/debian/scripts/init.sh#L30, will create the cache on container start. @turbo124 either we need this change to ensure that a bind mount has correct permissions or and I think this is the correct approach, remove https://github.com/invoiceninja/dockerfiles/blob/39c546e270f58e264ea6c5acdd7e084a1771908f/debian/docker-compose.yml#L20.

benbrummer:
@turbo124 the Docker guide for laravel `bootstrap/cache` is not persistent, only storage and in our case public needs to be handled for nginx. Do you agree on removing the `bootsrap/cache` volume?

https://docs.docker.com/guides/frameworks/laravel/ and volumes https://github.com/dockersamples/laravel-docker-examples/blob/main/compose.prod.yaml#L14

turbo124:
@benbrummer makes sense!

benbrummer:
@mindovermiles262 thanks for your contribution, we will remove the bootsrap/cache volume with https://github.com/invoiceninja/dockerfiles/pull/745

PR #761, opened 2025-06-10
New Alpine folder with Custom Dockerfile for new Alpine Image

charlesgek:
Hi ! I Forked the Dockerfile Repo, and worked on some optimisations for the Debian image, as well as an Alpine version with a patch allowing the installation and update of SaxonCHE (Open source version) bringing full compatibility for this image. Missing fonts are also included, copied from the Debian packages.

Debian or not, Saxon library is not compatible with x86 (i386) it is only compatible with x86_64 (amd64) and aarch64. This can be confirmed by looking at the code of the script or directly on the page of Saxonica :

https://www.saxonica.com/download/c.xml

I also made few changes :

    Google Chrome replaced by Chromium
    MySQL replaced by MariaDB (both images and compose file)
    Redis replaced by Valkey
    Included the missing fonts on Alpine, extracted from Debian packages

These changes solve inherent Open Source licensing issues and also bring optimizations (specially the change from Redis to Valkey).

I tested everything and it is working like a charm.

As information, I noticed the Debian image ended up with 2.68Go and the Alpine image with 2.16Go (After running the build squashing all the layers into one for production images)

### DIFF
--- a/debian/.env
+++ b/debian/.env
@@ -1,0 +1,62 @@
+# IN application vars
+APP_URL=http://localhost:8012
+APP_KEY=base64:RR++yx2rJ9kdxbdh3+AmbHLDQu+Q76i++co9Y8ybbno=
+APP_ENV=production
+APP_DEBUG=true
+REQUIRE_HTTPS=false
+PHANTOMJS_PDF_GENERATION=false
+PDF_GENERATOR=snappdf
+TRUSTED_PROXIES='*'
+
+
+CACHE_DRIVER=redis
+QUEUE_CONNECTION=redis
+SESSION_DRIVER=redis
+
+REDIS_HOST=redis
+REDIS_PASSWORD=null
+REDIS_PORT=6379
+
+FILESYSTEM_DISK=debian_docker
+
+# DB connection
+DB_HOST=mysql
+DB_PORT=3306
+DB_DATABASE=ninja
+DB_USERNAME=ninja
+DB_PASSWORD=ninja
+DB_ROOT_PASSWORD=ninjaAdm1nPassword
+DB_CONNECTION=mysql
+
+# Create initial user
+# Default to these values if empty
+IN_USER_EMAIL=admin@example.com
+IN_PASSWORD=changeme!
+# IN_USER_EMAIL=
+# IN_PASSWORD=
+
+# Mail options
+MAIL_MAILER=log
+MAIL_HOST=smtp.mailtrap.io
+MAIL_PORT=2525
+MAIL_USERNAME=null
+MAIL_PASSWORD=null
+MAIL_ENCRYPTION=null
+MAIL_FROM_ADDRESS='user@example.com'
+MAIL_FROM_NAME='Self Hosted User'
+
+# MySQL
+MYSQL_ROOT_PASSWORD=ninjaAdm1nPassword
+MYSQL_USER=ninja
+MYSQL_PASSWORD=ninja
+MYSQL_DATABASE=ninja
+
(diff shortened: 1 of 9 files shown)


turbo124:
@benbrummer thoughts? I'm not sure the juice is worth the squeeze here.

charlesgek:
Hi @turbo124 @benbrummer  ! I noticed on the Dockerfile repo that a backup script was missing to make.

I wanted to make a container that would do regular backups, it could do mysql or mariadb backup without issue and send it to the preferred backup solution. But since I know other things could be needed like logo or other things, I wondered which files or directory should be backed up ?

That way I can contribute to the repo as well.

Note : @hillelcoren suggested to have a look at  https://github.com/DrDBanner/inmanage on discord, I'll probably adapt this script to work with an additional container and add functions to send the backup to desired locations and send notification, email and/or webhook

benbrummer:
- MariaDB instead of MySQL is a good choice but a breaking change for existing users 
- valkey instead of redis is a good choice
- @turbo124 prefers google chrome over chromium
- saxonc compilation is interesting, even for debian, as saxonica did not release the libsaxon for 12.6 and 12.7

- Before merging this a general discussion about the handling and building of the images could make more sense. => alpine, debian, frankenphp https://thephp.foundation/blog/2025/05/15/frankenphp/, octane, supervisor vs multiple images, docker buildx, main-branch for images, git tags, build images automatically for each release...

- Alpine the base image size is small, but in relation to the size added by invoiceninja and dependencies is not really relevant

- Alpine is using musl, which falls behind libc in terms of PHP performance.

benbrummer:
> Hi @turbo124 @benbrummer  ! I noticed on the Dockerfile repo that a backup script was missing to make.
> 
> I wanted to make a container that would do regular backups, it could do mysql or mariadb backup without issue and send it to the preferred backup solution. But since I know other things could be needed like logo or other things, I wondered which files or directory should be backed up ?
> 
> That way I can contribute to the repo as well.
> 
> Note : @hillelcoren suggested to have a look at  https://github.com/DrDBanner/inmanage on discord, I'll probably adapt this script to work with an additional container and add functions to send the backup to desired locations and send notification, email and/or webhook

In my opinion backup is an individual thing. backup volumes, dump the database, backup a complete VM..., so I think a documentation + examples makes more sense.

benbrummer:
Thanks for your contribution 😀

charlesgek:
Hi ! @benbrummer thanks for the comments !

My script detects the latest version available of SaxonCHE from their website (12.7), download it and compile it. To use it on Debian, you just have to remove the line linked to the patch, it's just an adaptation to allow it to compile on Alpine.

Actually for MariaDB, if you keep the same Disk name in the .ENV  and keep the same volume name as the one previously used by MySQL, it doesn't break as both stores database at the same location. You can run some test to confirm 😁

For Google Chrome, I can understand, but it may pose some licensing issues down the line.

Everything else, I agree completely. Just a little note that Alpine is more streamlined and comes with less packages out of the box, which means installing the dependencies is necessary, and even musl being less performant with PHP, the fact that less processes are running, it actually increases the performances compared to the Debian image.

I also found this image which is basically a fork of alpine using glibc instead of musl, not sure what it's worth but I'll test it :
https://woahbase.online/images/alpine-glibc/

I also found an alternative solution, minideb from bitnami, which gives a very minimal Debian image (38Mo) and that would solve all the issues : 
https://github.com/bitnami/minideb

Finally for the backup, I was thinking of database backup and restoration and eventually additional files like images/logo for the invoices. This could facilitate migrations.

PR #762, opened 2025-06-11
New optimized Debian Image + Second folder OSS for fully Open Source

charlesgek:
Hi @turbo124 @benbrummer !
Following your advice yesterday, I went back to work on these.

Final changes : 
- Alpine put on hold and removed from the pull request
- Second folder debian-oss which is completely open source 
- Named volumes to avoid breaking changes (Mariadb and MySQL uses the same folder and are exchangeable)
- Named Containers to avoid future breaking changes
- Optimized Dockerfiles to reduce the number of steps and layers (and final size of the images)
- Redis replaced by Valkey for performances
- Script used via a curl function to install the latest version of saxon (12.7) 

If you are interested on the script, I set it on MIT Licence on my repo : 
https://github.com/charlesgek/SaxonCHE-PHP

There may be some minor other things I forgot, but these are the most important.


turbo124:
This approach I certainly support, we can add updated documentation to make it clear for users to select the appropriate folder they wish to start up the containers from. Will wait for review from @benbrummer

charlesgek:
Hi @turbo124 @benbrummer Hope you're both going well. 
Any update ?

benbrummer:
- I would at first just add debian-oss as new variant.
- The way of building https://raw.githubusercontent.com/charlesgek/SaxonCHE-PHP/refs/heads/main/Debian/saxon-install.sh could be contributedt to https://github.com/mlocati/docker-php-extension-installer => https://github.com/mlocati/docker-php-extension-installer/blob/d6ea5fbddfe5be579283ec6be21e3f6b02fe71e1/install-php-extensions#L3979, if EE can be replaced with HE
- We should reach out to Saxonica, if they can provide libsaxon-* again

charlesgek:
Hi @benbrummer about the download and install script, I sent a message a mlocati and went trough his script, but while I didn't receive any reply, his script would require significant changes, since I put my script on MIT licence, you can add it to your repo and just modify the section which select the latest version to instead download the one you want by pinning to a specific version with a download link instead of the detection.

For the other parts, I let you manage.

debian/Dockerfile
```diff
@@ -18,11 +18,21 @@ FROM php:${PHP}-fpm
...
+ARG php_suggest="exif imagick intl pcntl soap"
 ARG php_extra="opcache"
 
-# Install system dependencies
-RUN apt-get update && apt-get install -y --no-install-recommends \
+# Download script to install PHP extensions
+COPY --from=ghcr.io/mlocati/php-extension-installer /usr/bin/install-php-extensions /usr/local/bin/
+
+# Install the latest available version of SaxonCHE (Open Source) PHP Extension
+RUN ( curl -f -sSL https://raw.githubusercontent.com/charlesgek/SaxonCHE-PHP/refs/heads/main/Debian/saxon-install.sh || echo 'echo "Failed to download saxon-install.sh"; exit 1' ) | sh -s \
```
benbrummer, line 28: Preferable would be a integration into https://github.com/mlocati/docker-php-extension-installer, which is mentioned in the documentation for the official php docker images https://hub.docker.com/_/php, to make it available in the common way. But it is up to @turbo124 to decide on this.

PR #765, opened 2025-07-03
Update init.sh

stavros-k:
I believe this should be like that, otherwise this error appears
```sh
In Compiler.php line 67:

                                      

  Please provide a valid cache path.
```

### DIFF
--- a/debian/scripts/init.sh
+++ b/debian/scripts/init.sh
@@ -46,9 +46,9 @@ if [ "$*" = 'frankenphp php-cli artisan octane:frankenphp' ] || [ "${1#-}" != "$
 
         # Check for required folders and create if needed, relevant for bind mounts
         # It is not possible to chown, as we are not executing this script as root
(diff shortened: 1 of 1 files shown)


turbo124:
mmm I do not think this looks right.

There is no storage folder under the app/ folder path.

The framework files are all located under the storage/ folder path.

stavros-k:
> There is no storage folder under the app/ folder path.

The new volume appears to be under /app/storage
https://github.com/invoiceninja/dockerfiles/blob/f80e7f2ec3ed5e6ea4f26bfcf96ac27a3dedb097/debian/docker-compose.yml#L12
And I also see the app is copied under /app as well
https://github.com/invoiceninja/dockerfiles/blob/f80e7f2ec3ed5e6ea4f26bfcf96ac27a3dedb097/debian/Dockerfile#L85

turbo124:
@stavros-k thanks for this. I believe you are right!

@benbrummer i'm not sure which PR lead to the regression but .8 doesn't come up because the old paths cannot be cleared.

PR #770, opened 2025-08-01
Use default supervisord config as baseline

TheNewSound:
Start supervisord using /etc/supervisor/supervisord.conf so that a socket at /var/run is created (default supervisord config). This fixes the error message when trying to use supervisorctl inside the container.

Problem: In the current setup, supervisor is started with /etc/supervisor/conf.d/supervisord.conf, this ignores all default settings in /etc/supervisor/supervisord.conf (like the socket creation at /var/run, such that supervisorctl is able to access supervisor).

It is not needed to start supervisor with /etc/supervisor/conf.d/supervisord.conf instead of /etc/supervisor/supervisord.conf because the default config will load all configs at conf.d/*.conf

### DIFF
--- a/debian/Dockerfile
+++ b/debian/Dockerfile
@@ -41,7 +41,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && apt-get install -y --no-install-recommends google-chrome-stable; \
     elif [ "$(dpkg --print-architecture)" = "arm64" ]; then \
     apt-get install -y --no-install-recommends \
-    chromium; \        
+    chromium; \
     fi \
     # Create config directory for chromium/google-chrome-stable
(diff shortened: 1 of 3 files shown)


turbo124:
@benbrummer 

This looks reasonable? Thoughts?

benbrummer:
Indeed a good change to follow the convention with separate files for applications managed by supervisord.

turbo124:
Thanks!

turbo124:
@TheNewSound are you able to PR this to octane branch also please?

TheNewSound:
> @TheNewSound are you able to PR this to octane branch also please?

Octane branch/dockerfiles does not seem to use supervisord, but uses separate frankenphp containers for running the scheduler, queue and app

PR #777, opened 2025-08-23
feat(debian): add docker-compose.override.yml for optional mariadb setup; move .env to .env.example

nitanmarcel:
mysql seems to be death: `ERROR 2026 (HY000): TLS/SSL error: self-signed certificate in certificate c `


benbrummer:
Updating from mysql 8 to mariadb can only be done trough a database dump, so it affects all users, which directly run invoiceninja based on this repo.

nitanmarcel:
> Updating from mysql 8 to mariadb can only be done trough a database dump, so it affects all users, which directly run invoiceninja based on this repo.

Oh true, I forgot about already existing users. Oops

benbrummer:
@turbo124 I think it would make sense to have sample files like example.compose.yaml and example.env and add docker-compose.yaml and .env to .gitignore.
So no one has to handle conflicts and maintaining the compose.yaml and .env-file can be handled by the user. Afterwards it would be easier to make (breaking) changes in the example.compose.yaml, without causing troubles for anyone.

Example: https://github.com/benbrummer/dockerfiles/tree/octane-action

nitanmarcel:
> @turbo124 I think it would make sense to have sample files like example.compose.yaml and example.env and add docker-compose.yaml and .env to .gitignore. So no one has to handle conflicts and maintaining the compose.yaml and .env-file can be handled by the user. Afterwards it would be easier to make (breaking) changes in the example.compose.yaml, without causing troubles for anyone.
> 
> Example: https://github.com/benbrummer/dockerfiles

For the .env yeah that should have been the default, for easier updates.

Docker compose support merging via an override file that can be used.

https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/

nitanmarcel:
> @turbo124 I think it would make sense to have sample files like example.compose.yaml and example.env and add docker-compose.yaml and .env to .gitignore. So no one has to handle conflicts and maintaining the compose.yaml and .env-file can be handled by the user. Afterwards it would be easier to make (breaking) changes in the example.compose.yaml, without causing troubles for anyone.
> 
> Example: https://github.com/benbrummer/dockerfiles/tree/octane-action

it should be properly done now. with an override for compose to use mariadb if needed. Also with the `.env` to `.env.example` change

debian/docker-compose.override.yml.example
benbrummer, line 1: The override should use 'MARIADB_*' variables

Sidenote, a blankline at the end of files is recommended
image

debian/docker-compose.override.yml.example
```diff
@@ -0,0 +1,24 @@
...
+      MYSQL_DATABASE: ${DB_DATABASE}
+      MYSQL_USER: ${DB_USERNAME}
+      MYSQL_PASSWORD: ${DB_PASSWORD}
+      MYSQL_ROOT_PASSWORD: ${DB_ROOT_PASSWORD}
+    volumes:
+      - mariadb_data:/var/lib/mysql
+    networks:
+      - app-network
+    healthcheck:
+      test: [ "CMD", "mysqladmin", "ping", "-h", "localhost", "-u${MYSQL_USER}", "-p${MYSQL_PASSWORD}" ]
```
benbrummer, line 17: Healthcheck for mariadb `["CMD", "healthcheck.sh", "--connect", "--innodb_initialized"]`

benbrummer (commented):
Overwrite works as long as no one wants to rename the service. Same for redis/valkey overwrite. I guess we will have some people trying to get rid of mysql or redis where possible. But in general the overwrite for compose is a good way to allow modifications.

debian/docker-compose.override.yml.example
nitanmarcel: > The override should use 'MARIADB_*' variables
> 
> Sidenote, a blankline at the end of files is recommended image

Nope. MYSQL_* is good

https://mariadb.com/docs/server/server-management/install-and-upgrade-mariadb/configuring-mariadb/mariadb-environment-variables

debian/docker-compose.override.yml.example
nitanmarcel: > Healthcheck for mariadb `["CMD", "healthcheck.sh", "--connect", "--innodb_initialized"]`

mariadb is backwards compatible so it doesn't really matter

https://mariadb.com/docs/server/clients-and-utilities/legacy-clients-and-utilities/mysqladmin

benbrummer:
I would prefere a clean overwrite, where possible. healthcheck.sh would be in my opinion a better solution, which is only available for MariaDB. Yes indeed MariaDB is still somehow compatible to MySQL, but distros are getting rid of aliases, as this is changing (e.g Alpine). Therefore to be futureproof, I would suggest to use mariadb specific features and names, where possible. Why did you decide to add a new volume, the mysql volume would work, too ;-)

nitanmarcel:
> I would prefere a clean overwrite, where possible. healthcheck.sh would be in my opinion a better solution, which is only available for MariaDB. Yes indeed MariaDB is still somehow compatible to MySQL, but distros are getting rid of aliases, as this is changing (e.g Alpine). Therefore to be futureproof, I would suggest to use mariadb specific features and names, where possible. Why did you decide to add a new volume, the mysql volume would work, too ;-)

It failed to work for me

```
ERROR 2026 (HY000): TLS/SSL error: self-signed certificate in certificate c 
```

Plus mariadb is also open sourced under the GPL license and it's a little faster than mysql in some cases, specially when handling big data

benbrummer:
Now it is up to @turbo124 to decide on this.

nitanmarcel:
> I would prefere a clean overwrite, where possible. healthcheck.sh would be in my opinion a better solution, which is only available for MariaDB. Yes indeed MariaDB is still somehow compatible to MySQL, but distros are getting rid of aliases, as this is changing (e.g Alpine). Therefore to be futureproof, I would suggest to use mariadb specific features and names, where possible. Why did you decide to add a new volume, the mysql volume would work, too ;-)

Oh, I understood this wrongly. No idea, it just made sense to use a mariadb named volume to avoid any issues in the future

benbrummer:
> > I would prefere a clean overwrite, where possible. healthcheck.sh would be in my opinion a better solution, which is only available for MariaDB. Yes indeed MariaDB is still somehow compatible to MySQL, but distros are getting rid of aliases, as this is changing (e.g Alpine). Therefore to be futureproof, I would suggest to use mariadb specific features and names, where possible. Why did you decide to add a new volume, the mysql volume would work, too ;-)
> 
> Oh, I understood this wrongly. No idea, it just made sense to use a mariadb named volume to avoid any issues in the future

So the same idea as behind using MARIADB variables.

Thanks for your contribution!

Issue #780, opened 2025-08-27
`composer` inside docker container in the application's directory crashes.


TheNewSound:
**Describe the bug**
When running / executing the composer binary inside the InvoiceNinja docker container, it throws an error message:
```
PHP Fatal error:  Declaration of Symfony\Component\Console\Command\HelpCommand::execute(Symfony\Component\Console\Input\InputInterface $input, Symfony\Component\Console\Output\OutputInterface $output) must be compatible with Symfony\Component\Console\Command\Command::execute(Symfony\Component\Console\Input\InputInterface $input, Symfony\Component\Console\Output\OutputInterface $output): int in phar:///usr/local/bin/composer/vendor/symfony/console/Command/HelpCommand.php on line 70
``` 
This problem may be unrelated to the actual invoiceninja build, however its worth to note that the execution of `composer` inside the container also does not work for other directories than the application directory at `/var/www/html`. After some further digging it also turns out that the `scripts:` section of the invoiceninja composer.json is triggering this specific error. When I remove all members of the `scripts:` section after the dump autoload command, the `composer --version` command works again. But other `composer` commands still throws errors.

**To reproduce**
1. build the invoiceninja container using this dockerfile.
2. try to execute the `composer` command inside the container via a terminal.
3. error message appears.

**Expected behavior**
I should be able to use the composer command inside the docker container.

**Related issue**
https://github.com/composer/composer/issues/12514

**Solution**
Disable opcache in cli

PR #783, opened 2025-08-28
Update php.ini to disable opcache in cli

TheNewSound:
using `opcache.preload=/var/www/html/preload.php` and `opcache.enable_cli=1` together in php.ini is incompatible with composer and causes `composer` to throw error messages regarding the inclusion of wrong vendor/symfony files.

https://github.com/invoiceninja/dockerfiles/issues/780 https://github.com/composer/composer/issues/12514

### DIFF
--- a/debian/php/php.ini
+++ b/debian/php/php.ini
@@ -5,7 +5,7 @@ upload_max_filesize=10M
 
 [opcache]
 ; https://www.php.net/manual/en/opcache.installation.php#opcache.installation.recommended
-opcache.enable_cli=1
+opcache.enable_cli=0
 
 [jit]
 ; https://wiki.php.net/rfc/jit_config_defaults


turbo124:
These are only warnings, composer isn't available in the container?

cc @benbrummer

TheNewSound:
> These are only warnings, composer isn't available in the container?
> 
> cc @benbrummer

For me they are not just warnings, they are errors and execution of the `composer` executable is impossible

turbo124:
How are you running the application? the container does not include include composer at all.

TheNewSound:
Ah ye, I've added @composer to the php extensions in the dockerfile because I'm extending invoiceninja with custom code which are packages, e.g. migration files.

So yes indeed I'm running a custom dockerfile. So maybe this does not need to be merged in for this repo, however it would save many headaches for other people trying to achieve the same thing, or those running other PHAR's in the container.

benbrummer:
@TheNewSound, what about running composer with `php -d opcache.enable_cli=0`, same for preload `-d opcache.preload=''`?

TheNewSound:
> @TheNewSound, what about running composer with `php -d opcache.enable_cli=0`, same for preload `-d opcache.preload=''`?

Ye sure that works, but the point here is not to evade this problem once, but solve it so others will not stumble into this problem. The error message that people can stumble into makes it extremely hard for people to diagnose the root cause. And it can occur for any phar / php executable ran inside the container. 

There is also not really a performance benefit in opcache preloading for CLI:
https://github.com/composer/composer/issues/12514#issuecomment-3234322568
So in my opinion it is better to disable it so people can run their PHAR's / php executables without trouble inside the default dockerfile.

If you think disabling this for CLI in the php.ini is not ideal, I would propose to add a comment to the php.ini file near opcache.enable_cli that mentions this problem.

benbrummer:
I would say in production `opcache.enable_cli=1` is the way to go, especially for the cli commands for scheduler and worker. During development I would not enable opcache at all, only if something needs to be implemented/debugged for opcache e.g. preload.php.
For the compose use case a multistage docker build, as we use it could solve your issue, as the production deplyoment php settings are than out of your way.

Sidenote
Maybe https://github.com/php/php-src/blob/master/php.ini-development instead of the https://github.com/php/php-src/blob/master/php.ini-production, which we use in this image, can be usefull for you.

PR #789, opened 2025-09-01
Workaround: Disable SSL for mariadb-client for compatibility with MySQL

benbrummer:
Disable ssl in Dockerfile instead of init.sh, to align with octane


turbo124:
Thanks!

Issue #776, opened 2025-08-22
Bitnami images


sn0b4ll:
Hey Team,

First of, many thanks for this project and the charts!

I noticed that the helm charts currently strongly rely on bitnami images. With bitnami switching to an payed model, are there any plans to migrate the charts?

Greetings,
sn0b4ll

turbo124:
We rely on the community to update the charts. Also, this particular image is deprecated, so I do not envisage them being updated anytime soon.

If charts were to be recreated, I'd suggest doing so on the debian branch.

93imagine:
Where can I find the official community helm chart?

benbrummer:
> Where can I find the official community helm chart?

Only available on the deprecated master branch https://github.com/invoiceninja/dockerfiles/tree/master/charts/invoiceninja

Issue #763, opened 2025-06-21
Invoice ninja chart broken


FML128:
**Setup information**

Helm

**Describe the bug**

**To reproduce**
1. Install chart
2. Failure

**Expected behavior**
It should work

**Screenshots/logs**

```
  Normal   Pulling           7s (x3 over 47s)   kubelet            Pulling image "docker.io/bitnami/bitnami-shell:10"
  Warning  Failed            6s (x3 over 47s)   kubelet            Failed to pull image "docker.io/bitnami/bitnami-shell:10": failed to pull and unpack image "docker.io/bitnami/bitnami-shell:10": failed to resolve reference "docker.io/bitnami/bitnami-shell:10": pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
```

**Additional context**
Looks like the dependencies are finally to old and are beeing pulled. A update to the newsest release is needed.
Also, since it looks like the chart is not updated regularly it would make sense to set the defaults to rolling images to avoid security issues with older versions of the software. Or better regular updates to the chart

FML128:
This issue is not resolved yet fyi.

irgendwr:
I tried fixing the outdated dependencies, but didn't manage to get it working yet...
https://github.com/invoiceninja/dockerfiles/compare/master...irgendwr:invoiceninja-dockerfiles:master

Something with redis doesn't seem to work since I get the following error:
> production.ERROR: No sentinel server available for autodiscovery. {"exception":"[object] (Predis\\ClientException(code: 0): No sentinel server available for autodiscovery. at /var/www/app/vendor/predis/predis/src/Connection/Replication/SentinelReplication.php:302)

irgendwr:
It should work now with the patches in my fork.

But since it still uses bitnami images I would not recommend it (they're outdated and will eventually break). And this chart is unmaintained anyway.

(also I accidentally nuked my database in the process 😓  and this was quite annoying to investigate, so I'll not continue using this chart)

PR #801, opened 2025-09-18
Fix helm chart

irgendwr:
Related issues: #763, #776

This is an attempt at fixing the chart. More work is needed to actually remove the dependency on the deprecated bitnami images and charts.

Please verify this before merging.

I don't plan to continue working on this. This was already annoying enough.


irgendwr:
I'm completely fine with this being rejected, but automatically closing issues and pull-request after "inactivity" is not only annoying but feels very insulting after spending hours looking into it and submitting a patch. I don't think I will ever try to contribute here again.

turbo124:
Hi There, given the comment in the initial contribution:

```
I don't plan to continue working on this. This was already annoying enough.
```

We couldn't merge something that wouldn't be supported going forward - even for a small duration until we could get confirmation that this works. 

Thank you for PR'ing thou, If someone else is able to take what you have PR'd here are support it going forward we'd certainly look to merge this.

irgendwr:
Thanks for your reply! This is completely understandable and I don't take any issue with it.

My frustration is purely related to the automated closing of issues an PRs. I strongly suggest reconsidering this.

PR #805, opened 2025-09-26
Optimize artisan commands and add healthcheck as condition for app

### LINKED ISSUE #806: Correct order of artisan commands
@turbo124 I try to understand the order of the artisan commands.
@hillelcoren fix is almost ready

Current implementation
```
php artisan optimize
php artisan package:discover
php artisan migrate --force
...
php artisan db:seed --force
php artisan ninja:create-account
```
Result:
Localization dropdown incomplete, the optimization is populating the caches.
During the migration some things are generated but, than missing in the already populated cache

Fix or workaround:
Just run `php artisan config:cache` in the first place and `php artisan optimize` after the `php artisan migrate --force`, alternative is to run "php artisan optimize" twice, but `php artisan cache:clear` before the second call.

Option 1
```
php artisan config:cache
php artisan package:discover
php artisan migrate --force
php artisan optimize
...
php artisan db:seed --force
php artisan ninja:create-account
...
```

Option 2
```
php artisan optimize
php artisan package:discover
php artisan migrate --force
php artisan optimize:clear // php artisan cache:clear
php artisan optimize
...
php artisan db:seed --force
php artisan ninja:create-account
...
```

Question:
Can we avoid populating caches multiple times, with the need to clear them in between? So do we have a better Option 3? https://laravel.com/docs/12.x/cache#removing-items-from-the-cache?

benbrummer:
Closes #806 
Make use of https://github.com/invoiceninja/dockerfiles/commit/1bda3e58094dc83ddd064a6a59c2650eb18f4563


benbrummer:
image

turbo124:
Thanks!

PR #808, opened 2025-09-29
Remove artisan package:discover

benbrummer:
- Files are already part of invoiceninja.tar.gz
- Upgrading can be done without bringing the stack down. Afterwards docker compose up -d will only recreate the container if an update has been pulled


turbo124:
Thanks!

PR #813, opened 2025-09-30
Push README.md to Dockerhub

benbrummer:
- Push README.md as description to https://hub.docker.com/r/invoiceninja/invoiceninja-octane
- Use default depth (1) for checkout


benbrummer:
@turbo124 should we do the same for invoiceninja-debian? The current instructions are not clear at https://hub.docker.com/r/invoiceninja/invoiceninja-debian and I guess the README.md is much better. Minor issue are the README.md emojies, which are only displayed peroperly on GitHub, so we might use some standardized onces.

turbo124:
@benbrummer sure thing. I'll follow your lead! :)

PR #816, opened 2025-10-01
Compose file refactoring - Debian

benbrummer:
* .env file was passing all (secrets) to mysql
* .env file volume and env_file: where used redundantly => volume was removed
* docker logging extension has been removed. docker by default will rotate and most logs are logged to laravel.log
* healthcheck timing has been removed. Default timings https://docs.docker.com/reference/dockerfile/#healthcheck are sufficient. For invoiceninja a specific timing is configured in dockerfile.


turbo124:
Looks great! Thanks!

PR #818, opened 2025-10-20
manifest creation for multiple runners

benbrummer:
Based on https://docs.docker.com/build/ci/github-actions/multi-platform/#build-and-load-multi-platform-images

push will not happen with `docker/build-push-action` but with `docker buildx imagetools create` for the release.

If there is a non-release build 'docker buildx imagetools create --dry-run', will be executed, which will not push

### DIFF
--- a/.github/workflows/publish-image.yaml
+++ b/.github/workflows/publish-image.yaml
@@ -10,12 +10,15 @@ env:
 
 jobs:
   build:
-    runs-on: ${{ startsWith(matrix.platforms, 'linux/arm') && 'ubuntu-24.04-arm' || 'ubuntu-latest' }}
+    runs-on: ${{ startsWith(matrix.platform, 'linux/arm') && 'ubuntu-24.04-arm' || 'ubuntu-latest' }}
+    outputs:
+      tags: ${{ steps.prep.outputs.tags }}
     strategy:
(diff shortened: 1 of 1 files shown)


benbrummer:
@turbo124 please review this way of building

It is following the example from docker.  https://docs.docker.com/build/ci/github-actions/multi-platform/#build-and-load-multi-platform-images

Build: Will push the images only with their digest to docker hub, so the runners do not overwrite each others
Merge: Will create a manifest to dockerhub referencing the previously pushed images. The argument --dry-run will prevent the push of this manifest

Benefit: Fast builds
Current downside: Dangling images on docker hub, but there seams to be something on roadmap for dockerhub to delete these.

Can you check if there is a different trigger configuration possible, which fits your needs? I think it is quite difficult to review this PR without a pipeline

turbo124:
@benbrummer It looks good to me, but i am certainly no expert.

We can preform some tagged releases to test this now if you like?

benbrummer:
I just pushed the updated publish-image.yaml. So testing can be done. But I need a break now :-)

https://www.dockerstatus.com/ I hope the disruptions will not make any (further) troubles

turbo124:
Thanks @benbrummer appreciate your help!!

PR #819, opened 2025-10-20
Rename secret

benbrummer:
Remark: Was it intended to use https://github.com/invoiceninja/dockerfiles/tree/refs/tags/5.11.28-d1 instead of 5.18.28-d1

### DIFF
--- a/.github/workflows/build-image-debian.yaml
+++ b/.github/workflows/build-image-debian.yaml
@@ -89,7 +89,7 @@ jobs:
         uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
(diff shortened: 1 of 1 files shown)


turbo124:
@benbrummer build is working!

The only thing that is not clear to me is whether any additional configuration is required for selecting the platform, or whether this is handled automagically?

benbrummer:
> @benbrummer build is working!
> 
> The only thing that is not clear to me is whether any additional configuration is required for selecting the platform, or whether this is handled automagically?

This is happening automatically. The build will use a matrix to select the correct runner amd64 or arm64. The manifest/merge job will get the digest of the images created on the "independent" runners and creates a manifest referencing both of them. For the user nothing changes, cause the os will pull the matching architecture as before.

turbo124:
Fantastic!

PR #829, opened 2025-10-29
url as argument

benbrummer:
- url is now derived from the tag for downloading a defined release.
- ADD sets permission 500 so extraction can not be done by an other user. Therefore chown is needed for ADD for a successful extraction by www-data


turbo124:
Thanks!

PR #868, opened 2026-02-25
Rollback rsync for debian

turbo124:


marcolaux:
this broke ZFS setups again

PR #873, opened 2026-04-06
Add a helm chart for K8S

eburgueno:
This PR adds support to deploy IN to a Kubernetes cluster using Helm.

The chart is based on the current `docker-compose.yml`, and allows controlling several aspects of the installed release, including storage persistence, release version, and security context. See `values.yaml` for the defaults.

Tested on K8S 1.35.x using both a local dev (Talos on docker) and production clusters (Talos on baremetal).

### DIFF
--- a/charts/invoiceninja/Chart.yaml
+++ b/charts/invoiceninja/Chart.yaml
@@ -1,0 +1,6 @@
+apiVersion: v2
+name: invoiceninja
+description: Helm chart for Invoice Ninja Debian stack (app, nginx, mysql, redis)
+type: application
+version: 0.1.0
+appVersion: "latest"
--- a/charts/invoiceninja/Chart.yaml
+++ b/charts/invoiceninja/Chart.yaml
@@ -3,4 +3,6 @@ name: invoiceninja
 description: Helm chart for Invoice Ninja Debian stack (app, nginx, mysql, redis)
 type: application
 version: 0.1.0
-appVersion: "latest"
+appVersion: "5"
+maintainers:
+  - name: eburgueno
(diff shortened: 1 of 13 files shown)


turbo124:
@benbrummer thoughts?

eburgueno:
I'm happy to work on updating the GitHub Actions as well, just wanted to suggest this first before I spend time on that :)

turbo124:
@eburgueno we appreciate the contribution! An updated helm chart has been a long time request so we would certainly appreciate this!

turbo124:
@benbrummer do you have any spare cycles to evaluate this one for merging?

benbrummer:
@turbo124 the only thbing what I would change is the folder structure
`/charts` instead of `/charts/invoiceninja`, apart of this I'm fine with the implementation, although I need to admit I'm no expert on this.

turbo124:
@eburgueno does the path change sound reasonable? also, if you'd like to push any changes requires for the GH Actions, that would be super cool too :+1:

eburgueno:
I understand the intention behind the path suggestion, but sadly helm is a bit restrictive when it comes to the files structure. The convention is to use the chart name as the top level dir (without the version) of the main `Chart.yaml` file: https://helm.sh/docs/topics/charts#the-chart-file-structure

Happy to work on the CI/CD part too. I'll try and get to that during the weekend.

turbo124:
@eburgueno thanks for the update.

Can you advise whether we should disable the current GH actions? I assume they won't work with the new changes?

turbo124:
Thanks!

PR #876, opened 2026-04-08
feat: backup script

gegabega:
## Summary

- Automatic daily/weekly/monthly backups via cron inside the app container
- Backs up MySQL (mysqldump) + storage volume into timestamped `.tar.gz` archives (format YYYY-MM-DD-<frequency>.tar.gz)
- Tiered retention: daily 7d, weekly 30d, monthly 120d (took inspiration from your old master branch here)
- Uses basename/symlink pattern (also similar to your old master)

## Changes

- `debian/scripts/backup.sh` – backup script with frequency detection via `basename`
- `debian/Dockerfile` – add `cron`, copy script, symlink to cron.daily/weekly/monthly
- `debian/supervisor/supervisord.conf` – add `[program:cron]`
- `debian/docker-compose.yml` – add `app_backups` volume
- `debian/scripts/init.sh` – ensure `/backups` dir on startup
- `README.md` – backup/restore docs, checked off todo
- `debian/scripts/test_backup.sh` – shell unit test **(can be removed if unwanted ofc)**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,9 +6,13 @@
 .Trashes
 ehthumbs.db
 Thumbs.db
+.env
 
 # Helm
 charts/**/charts/
 
 # Compose filesystem
 /docker
+
+# Backups
+debian/backups/
--- a/README.md
+++ b/README.md
@@ -80,6 +80,44 @@ docker compose up -d
 
 It is recommended to perform a backup before updating.
 
+### Backup and Restore
+
+A backup script is included at `scripts/backup.sh`. It dumps the MySQL database and archives the storage volume into a single timestamped `.tar.gz` file.
+
+**Run from the `debian/` directory** (where `docker-compose.yml` lives):
+
+```bash
+./scripts/backup.sh
+```
+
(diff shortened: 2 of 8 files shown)


benbrummer:
Looks like an interesting approach, but in my opinion backup is a highly individual topic.

- Some might want to have system level backups (Proxmox VE...)
- Some want to have docker volume backups
- Some want to have a cron job outside of the image
- Some use docker exec to dump the database following mysql/mariadb guides

So I think it is a nice approach, but I'm not sure how to handle this => @turbo124 ?

.gitignore
```diff
@@ -6,9 +6,13 @@
 .Trashes
 ehthumbs.db
 Thumbs.db
+.env
```
benbrummer, line 9: I like the Idea of having .env-files outside of git and provide https://github.com/invoiceninja/dockerfiles/pull/888

turbo124: @benbrummer I believe in the master branch, we just have the option to remove the comment on the backup script init that is in the docker-compose.yml file. 

Then it just becomes a case of "opt in" whenever a users wants to use the provided functionality.

PR #885, opened 2026-07-27
feat: make redis and mariadb accept external instances

Softyy:
I wanted to use external instances of redis and mariadb with your helm chart. Here are the changes to make it happen. I wasn't sure if you were taking PRs, feel free to close this if it's not wanted.

This guards the resources via `mysql.enabled` and `redis.enabled`, so they're not added if they're not wanted. Then swaps the secret refs to the new `secret.existingSecret`, if it's provided.

Example values I'm using for this:

```
...

mysql:
  enabled: false
redis:
  enabled: false

secret:
  existingSecret: invoice-ninja

env:
  appUrl: https://<host>
  requireHttps: "true"
  dbHost: mariadb.mariadb.svc
  dbDatabase: ninja
  dbUsername: <username>
  redisHost: redis-master.redis.svc

...
```

Things seem to be fine on my end, everything deployed correctly.

### DIFF
--- a/charts/invoiceninja/templates/_helpers.tpl
+++ b/charts/invoiceninja/templates/_helpers.tpl
@@ -43,3 +43,11 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- define "invoiceninja.redisServiceName" -}}
 {{- printf "%s-redis" (include "invoiceninja.fullname" .) -}}
 {{- end -}}
+
+{{- define "invoiceninja.secretName" -}}
+{{- if .Values.secret.existingSecret -}}
+{{- .Values.secret.existingSecret -}}
+{{- else -}}
+{{- printf "%s-secret" (include "invoiceninja.fullname" .) -}}
+{{- end -}}
+{{- end -}}
--- a/charts/invoiceninja/templates/app.yaml
+++ b/charts/invoiceninja/templates/app.yaml
@@ -34,14 +34,18 @@ spec:
             - /bin/sh
             - -c
             - |
+              {{- if .Values.mysql.enabled }}
               until nc -z {{ include "invoiceninja.mysqlServiceName" . }} {{ .Values.mysql.service.port }}; do
                 echo "waiting for mysql";
                 sleep 2;
               done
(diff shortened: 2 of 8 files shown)


benbrummer:
@eburgueno can you have a look on this PR?

eburgueno:
I looks safe to me, but I would add two changes, one of them mandatory:

1. `Chart.yaml` needs a version bump, so CI works as expected (mandatory)
2. The busybox init container's wait-loop is fully skipped when `mysql.enabled`/`redis.enabled` is false, even when an external host is configured. So in external mode, the app pod can start before the external MySQL/Redis is actually reachable. I would leave the wait-loop working also for external hosts so that users don't end up with a failed deploy and blame the chart for it. But that's more a design decision around user experience than a breaking change.

And a warning: in `configmap.yaml`, if someone sets `mysql.enabled: false` (or `redis.enabled: false`) but forgets to set `env.dbHost`/`env.redisHost`, the `DB_HOST`/`DB_PORT` (or `REDIS_HOST`/`REDIS_PORT`) keys are silently omitted from the ConfigMap entirely. You get the error when the app tries to run rather than from Helm when you try to deploy. I would suggest updating the templating or the docs, but again, this design decision around user experience and won't break existing deployments.

Side note: @benbrummer you'll need to approve and run workflows for this PR to trigger the Lint and Test action (which would've caught the version bump) since this PR comes from a fork and first-time/outside contributor.

turbo124:
@Softyy are you able to align this PR with the recommendations from @eburgueno ?

Softyy:
> @Softyy are you able to align this PR with the recommendations from @eburgueno ?

@turbo124 sure, I can make the changes tomorrow.

turbo124:
@Softyy thanks, no rush! Appreciate the help.

Softyy:
@turbo124 updates have been made.

1. bumped the chart version
2. left in the busybox checks
3. added a check to the values so helm will complain about missing external hosts.

turbo124:
@Softyy thanks!

@eburgueno  if you have some spare cycles could you please review?

PR #888, opened 2026-09-14
Rename .env to .env.example

lilchancep:
This change offers a couple of benefits. 

1. When a user does a git pull to do an update, git will no longer ask to stash changes since we are not changing a file that is tracked. 
2. Allows maintainers to ship a clean version of the .env file that can always be used for reference.


turbo124:
You shouldn't need to do a git pull once the application has been booted.

```bash
docker compose down
docker compose pull
docker compose up
```

The repo combines both the build infrastructure AND the docker-compose file which we probably should abstract out at some stage as they are different contexts. But I do not think adding this change is worthwhile as it just adds friction to the initial configuration of the docker setup.

lilchancep:
The Docker-Compose references files that are kept within the NGINX directory that is served by the repo, is it safe to assume that the maintainers will never change said files/the person running invoice ninja is responsible for any changes.

If that is the case then I can close this.

turbo124:
thanks!

PR #890, opened 2026-09-17
fix for pdf generation

benbrummer:
chrome/chromium require nowadays an additional config folder chromium-headless/google-chrome-headless, which was breaking pdf generation. To fix this and to ensure that appliucations can create the required config files the folder /config is now owned by ninja


turbo124 (approved):
Feel free to merge when ready :)

